PR #3, opened 2015-10-07
fix deprecation warnings

yeesian:


yeesian:
Sorry I messed up here (didn't check for compatibility with 0.3), and have reverted and force-pushed the changes back to master.

PR #4, opened 2015-10-07
WIP: fix deprecation warnings

yeesian:
I can't seem to get the `Union{..}` stuff to work

```julia
  | | |_| | | | (_| |  |  Version 0.3.11 (2015-07-27 06:18 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> using Compat

julia> @compat(Union{Int,Float64})
ERROR: type: instantiate_type: expected TypeConstructor, got Function
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.3
+Compat
--- a/src/GeoInterface.jl
+++ b/src/GeoInterface.jl
@@ -1,5 +1,7 @@
 module GeoInterface
 
+    using Compat
+
     export  AbstractPosition, Position,
             AbstractGeometry, AbstractGeometryCollection, GeometryCollection,
             AbstractPoint, Point,
--- a/src/geotypes.jl
+++ b/src/geotypes.jl
@@ -1,7 +1,7 @@
 # Coordinate Reference System Objects
 # (has keys "type" and "properties")
 # TODO: Handle full CRS spec
-typealias CRS Dict{String,Any}
+typealias CRS Dict{@compat(AbstractString),Any}
 
 # Bounding Boxes
 # The value of the bbox member must be a 2*n array,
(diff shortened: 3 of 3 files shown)


visr:
Works for me, is your Compat up to date? I believe you need at least Compat 0.7.1

yeesian:
can you give it (and `LibGEOS.jl` [if you have it]) a spin? Thanks!

visr:
Tests pass cleanly for GeoInterface on 0.3 and 0.4. Though it's probably better to add Compat to REQUIRE. For LibGEOS, I had tried the `bindeps` branch before, and it's working well over WinRPM, very convenient. Now testing the tagged LibGEOS I only get a warning from FactCheck, that updated it's syntax:
`WARNING: The `=>` syntax is deprecated, use `-->` instead`
And on 0.4 a few deprecation warnings of course.

yeesian:
I haven't touched LibGEOS in a while.. I have been thinking of replacing it with OGR instead, since OGR geometries "talk to" GDAL, whereas GEOS (by itself) doesn't

visr:
Strictly speaking it would indeed fit better with JuliaGeometry. Don't know how the OGR functionality compares to GEOS standalone. Haven't really used LibGEOS yet, but I have used [Shapely](http://toblerity.org/shapely/manual.html). There you have to use functions like shape and mapping to make the jump between the feature data and the geometries and back.

yeesian:
Sean's work on shapely and fiona and rasterio have definitely influenced me most (reflected in this package too), but I don't think it's necessarily a good idea to follow them all the way, since

- [the functionality of OGR in 2007](http://osdir.com/ml/gis.geos.devel/2007-03/msg00011.html) is quite different from the functionality today,
- they need to [make the distinction for prepared geometries](http://lists.osgeo.org/pipermail/gdal-dev/2013-April/036101.html)
- it seems like they need to make unnecessary copies when converting their shapely (geos) geometries to ogr geometries in rasterio. See
   - https://github.com/mapbox/rasterio/blob/6b02fd304d10995cff818729abe47f28bd7a33b5/rasterio/features.py#L309-L331, and
    - https://github.com/mapbox/rasterio/blob/dfe498eda09689ccf62d1c07e2205409bc2da99f/rasterio/_features.pyx#L267-L280

yeesian:
Which also means that both `GeoInterface` and `LibGEOS` will become less and less relevant to JuliaGeo. With our current libraries, we can't really do generic operations on arbitrary julia types, without constructing the corresponding OSGeo objects first. It's probably better to acknowledge that upfront, rather than take the penalty in object transformations over and over for users.

yeesian:
@garborg, okay to merge? Only LibGEOS depends on this package

garborg:
Looks good to me.

PR #6, opened 2016-08-03
drop julia 0.3 support

visr:
I think after merging this I'll tag a new version. The current release is quite old and uses `Nothing`, which was deprecated in v0.4, and is gone in v0.5.

Are there any other things that should still be done before tagging a new version?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ os:
   - linux
   - osx
 julia:
-  - release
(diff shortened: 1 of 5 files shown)


yeesian:
Thanks so much! Can you help doublecheck that the tests for LibGEOS passes as well? I imagine it should be okay since you've also dropped support for 0.3 in LibGEOS.

PR #8, opened 2017-04-26
Update for julia0.5

yeesian:


yeesian:
Since this is a breaking change, I might as well include the proposed change in #9 here too.

yeesian:
Can you review the latest commit (since it was made after your previous review)?

visr:
I added one more commit, looks good to go now.

PR #10, opened 2017-05-23
fix deprecation warnings on julia v0.6

visr:
fix abstract type with Compat
change typealias A B to const A = B


visr:
This is not very controversial, will merge and tag tomorrow if nobody objects or beats me to it.

PR #12, opened 2017-12-31
Add support for integration into Plots ecosystem

yeesian:
@visr @mkborregaard

Based on the discussion in https://github.com/JuliaGeo/Shapefile.jl/issues/12

So the example in the README of PlotRecipes.jl will now be:

```julia
using Plots, Shapefile
dir = "https://github.com/nvkelso/natural-earth-vector/raw/master/110m_physical/"
fn = "ne_110m_land.shp"
run(`wget $dir/$fn -P /tmp/`)
shp = open("/tmp/$fn") do fd
    read(fd, Shapefile.Handle)
end
plt = plot()
for s in shp.shapes; plot!(plt, s); end
plt
```
and yield
example

Suggestions/improvements welcome!

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -1,13 +1,25 @@
 shapecoords(geom::AbstractPoint) = [tuple(coordinates(geom)...)]
-RecipesBase.@recipe f(geom::AbstractPoint) = (seriestype --> :scatter; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractPoint) = (
+    aspect_ratio := 1;
+    seriestype --> :scatter;
+    legend --> :false;
+    shapecoords(geom)
+)
 
-shapecoords(geom::AbstractVector{<:AbstractPoint}) = Tuple{Float64,Float64}[tuple(coordinates(g)...) for g in geom]
+shapecoords(geom::AbstractVector{<:AbstractPoint}) = Tuple{Float64,Float64}[
+    tuple(coordinates(g)...) for g in geom
+]
 
 function shapecoords(geom::AbstractMultiPoint)
     coords = coordinates(geom)
     first.(coords), last.(coords)
 end
-RecipesBase.@recipe f(geom::AbstractMultiPoint) = (seriestype --> :scatter; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractMultiPoint) = (
+    aspect_ratio := 1;
+    seriestype --> :scatter;
+    legend --> :false;
+    shapecoords(geom)
+)
 
 function shapecoords(geom::Vector{<:AbstractMultiPoint})
     x = Float64[]; y = Float64[]
@@ -22,7 +34,12 @@ function shapecoords(geom::AbstractLineString)
     coords = coordinates(geom)
     first.(coords), last.(coords)
 end
-RecipesBase.@recipe f(geom::AbstractLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractLineString) = (
+    aspect_ratio := 1;
+    seriestype --> :path;
+    legend --> :false;
+    shapecoords(geom)
+)
 
 function shapecoords(geom::Vector{<:AbstractLineString})
     x = Vector{Float64}[]; y = Vector{Float64}[]
@@ -39,7 +56,12 @@ function shapecoords(geom::AbstractMultiLineString)
     y = Vector{Float64}[last.(line) for line in coords]
     x, y
 end
-RecipesBase.@recipe f(geom::AbstractMultiLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractMultiLineString) = (
+    aspect_ratio := 1;
+    seriestype --> :path;
+    legend --> :false;
+    shapecoords(geom)
+)
 
 function shapecoords(geom::Vector{<:AbstractMultiLineString})
     x = Vector{Float64}[]; y = Vector{Float64}[]
@@ -53,7 +75,12 @@ function shapecoords(geom::AbstractPolygon)
     ring = first(coordinates(geom)) # currently doesn't plot holes
     first.(ring), last.(ring)
 end
-RecipesBase.@recipe f(geom::AbstractPolygon) = (seriestype --> :shape; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractPolygon) = (
+    aspect_ratio := 1;
+    seriestype --> :shape;
+    legend --> :false;
+    shapecoords(geom)
+)
 
 function shapecoords(geom::Vector{<:AbstractPolygon})
     x = Vector{Float64}[]; y = Vector{Float64}[]
@@ -72,7 +99,12 @@ function shapecoords(geom::AbstractMultiPolygon)
     end
     x, y
 end
-RecipesBase.@recipe f(geom::AbstractMultiPolygon) = (seriestype --> :shape; legend --> :false; shapecoords(geom))
+RecipesBase.@recipe f(geom::AbstractMultiPolygon) = (
+    aspect_ratio := 1;
+    seriestype --> :shape;
+    legend --> :false;
+    shapecoords(geom)
+)
 
 function shapecoords(geom::Vector{<:AbstractMultiPolygon})
     x, y = Vector{Float64}[], Vector{Float64}[]
@@ -84,13 +116,14 @@ function shapecoords(geom::Vector{<:AbstractMultiPolygon})
 end
 
 RecipesBase.@recipe function f(geom::Vector{<:AbstractGeometry})
+    aspect_ratio := 1
     legend --> :false
     for g in geom
         @series begin
             if g isa AbstractPoint || g isa AbstractMultiPoint
                 seriestype := :scatter
             elseif g isa AbstractLineString || g isa AbstractMultiLineString
-                seriestype := :line
+                seriestype := :path
             elseif g isa AbstractPolygon || g isa AbstractMultiPolygon
                 seriestype := :shape
             end
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -27,9 +27,9 @@ RecipesBase.@recipe f(geom::AbstractMultiPoint) = (
 
 function shapecoords(geom::Vector{<:AbstractMultiPoint})
     x = Float64[]; y = Float64[]
-    for g in geom
-        coords = coordinates(g)
-        append!(x, first.(coords)); append!(y, last.(coords))
+    for g in geom, pt in coordinates(g)
+        push!(x, pt[1]); push!(x, NaN)
+        push!(y, pt[2]); push!(y, NaN)
     end
     x, y
 end
@@ -55,9 +55,8 @@ function shapecoords(geom::Vector{<:AbstractLineString})
 end
 
 function shapecoords(geom::AbstractMultiLineString)
-    coords = coordinates(geom)
     x, y = Float64[], Float64[]
-    for line in coords
+    for line in coordinates(geom)
         append!(x, first.(line)); push!(x, NaN)
         append!(y, last.(line)); push!(y, NaN)
     end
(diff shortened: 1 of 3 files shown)


src/plotrecipes.jl
```diff
@@ -0,0 +1,55 @@
...
+            push!(linex, xi); push!(liney, yi)
+        end
+        push!(x, linex); push!(y, liney)
+    end
+    x, y
+end
+RecipesBase.@recipe f(geom::AbstractMultiLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
+
+function shapecoords(geom::AbstractPolygon)
+    x, y = [], []
```
juliohm: Perhaps you could replace this function body with:

```julia
coords = first(coordinates(geom))

first.(coords), last.(coords)
```

which is type stable.

src/plotrecipes.jl
```diff
@@ -0,0 +1,55 @@
...
+    for (xi,yi) in first(coordinates(geom))
+        push!(x, xi); push!(y, yi)
+    end
+    x, y
+end
+RecipesBase.@recipe f(geom::AbstractPolygon) = (seriestype --> :shape; legend --> :false; shapecoords(geom))
+
+function shapecoords(geom::AbstractMultiPolygon)
+    x, y = [], []
```
juliohm: Similar comment here, maybe rewrite it so that the types of x and y don't switch from Any to Float64 during processing.

yeesian:
Thanks for the review! I'll leave the PR open for a little longer in case others have comments.

src/plotrecipes.jl
```diff
@@ -2,53 +2,35 @@ shapecoords(geom::AbstractPoint) = [tuple(coordinates(geom)...)]
 RecipesBase.@recipe f(geom::AbstractPoint) = (seriestype --> :scatter; legend --> :false; shapecoords(geom))
 
 function shapecoords(geom::AbstractMultiPoint)
-    x, y = [], []
-    for (xi,yi) in coordinates(geom)
-        push!(x, xi); push!(y, yi)
-    end
-    x, y
+    coords = coordinates(geom)
+    first.(coordinates(geom)), last.(coordinates(geom))
```
juliohm: Actually, I think you ment coords here, right? Almost missed it. :blush:

juliohm:
My pleasure, I am looking forward to see GeoInterface.jl everywhere in Geo-packages.

There is one more comment I made on the last commit.

mkborregaard:
The implementation currently here https://github.com/JuliaPlots/PlotRecipes.jl/blob/master/src/misc.jl#L5-L48

supports the syntax
```julia
plot(shp)
```
rather than
```julia
plt = plot()
for s in shp.shapes; plot!(plt, s); end
plt
```
I tend to prefer the short format - is there a reason to go for the long format here?

mkborregaard:
Using a series recipe would also simplify setting `aspect_ratio --> 1` which I think should be a default for spatial objects.

yeesian:
> I tend to prefer the short format - is there a reason to go for the long format here?

This is now possible with the vectorized commit (https://github.com/JuliaGeo/GeoInterface.jl/pull/12/commits/69cb0a5d3795bb6b577e635acccb27bd12c91c42); so you can run the following without having to define the recipe in Shapefile, or use PlotRecipes:

```julia
using Plots, Shapefile
dir = "https://github.com/nvkelso/natural-earth-vector/raw/master/110m_physical/"
fn = "ne_110m_land.shp"
run(`wget $dir/$fn -P /tmp/`)
shp = open("/tmp/$fn") do fd
    read(fd, Shapefile.Handle)
end
plot(shp.shapes)
```

Based on my understanding, these recipes only work for homogeneous collections of geometries, since points and lines and polygons have different `seriestype`. Do you have a proposal to make them work for arbitrary collections of geometries?

yeesian:
> Using a series recipe would also simplify setting `aspect_ratio --> 1` which I think should be a default for spatial objects.

I'm not sure what a series recipe is, but this can be done in other packages, e.g. https://github.com/yeesian/MapTiles.jl/blob/master/src/MapTiles.jl#L13-L19 without having to be enforced here?

So something like
```julia
import MapTiles, GeoJSON
using Plots

# ;wget https://raw.githubusercontent.com/awoodruff/leaflet-intro/gh-pages/neighborhoods.geojson
# ;wget https://raw.githubusercontent.com/awoodruff/leaflet-intro/gh-pages/rodents.geojson

nbhd = GeoJSON.parsefile("neighborhoods.geojson");
rodents = GeoJSON.parsefile("rodents.geojson");

basemap = MapTiles.fetchmap(
    -71.2, # minlon
    42.25, # minlat
    -71, # maxlon
    42.41, # maxlat
    provider = MapTiles.StamenProvider()
)
plt = plot(basemap)
for f in nbhd.features
    plot!(plt, MapTiles.project(basemap, f.geometry), color="cornflowerblue", alpha=0.5)
end
for f in rodents.features
    plot!(plt, MapTiles.project(basemap, f.geometry), color="orange", alpha=0.5)
end
plt
```
should work.

mkborregaard:
Yes - invent a `spatialpolygon`, `spatialline`, and `spatialpoint` seriestype with a series recipe, and set that inside a `@series` block.

mkborregaard:
Yes that should be possible - I'd need to play with the codes a bit to see where to insert it but I'm happy to have a look.

yeesian:
> Yes - invent a spatialpolygon, spatialline, and spatialpoint seriestype with a series recipe, and set that inside a @series block.

Is that a `PlotRecipes` thing, or a `GeoInterface` thing? Maybe I'll understand it better if I see an example

mkborregaard:
So to be clear I suggest scrapping the PlotRecipes implementation once this is merged - it's a much nicer experience for users to have it directly here.
There's an example of a series recipe here https://github.com/JuliaPlots/PlotRecipes.jl/blob/master/src/misc.jl#L6-L14 , which is just something that defines a new seriestype. So what I suggest is to use a user recipe (which is something that defines a type dispatch for plot) to split geometries into parts, then use `@series` to plot each part as a new series. That makes it easy to interactively add further series.

yeesian:
> So what I suggest is to use a user recipe (which is something that defines a type dispatch for plot) to split geometries into parts, then use at-series to plot each part as a new series. That makes it easy to interactively add further series.

Is this in response to: "Based on my understanding, these recipes only work for homogeneous collections of geometries, since points and lines and polygons have different seriestype. Do you have a proposal to make them work for arbitrary collections of geometries?"

I'll be happy if that's possible. But I can't yet figure out the corresponding recipe to write for that to happen, and will appreciate some help with it. I'll be happy to discuss further in slack.

mkborregaard:
Here's a dummy example, maybe that makes it clearer:
```julia
using Plots

abstract type MyTypes end
struct MyPoint <: MyTypes end
struct MyLine <: MyTypes end

struct MyHandle
    a::Vector{MyTypes}
end

@recipe function f(h::MyHandle)
    for t in h.a
        @series begin
            if t isa MyPoint
                seriestype := :scatter
            elseif t isa MyLine
                seriestype := :path
            else
                error("undefined MyTypes object")
            end
            rand(10)
        end
    end
end

obj = MyHandle([MyPoint(), MyLine()])
plot(obj)
```

juliohm:
@yeesian, as a side question, could you please explain how MapTiles.jl compare with OpenStreetMap.jl? The idea is to have a Plots.jl-compatible solution?

yeesian:
MapTiles fetches rastertiles that you can "stitch" together to form a basemap (mosaic of tiles), that you can plot on top of. Yes, they are going to be Plots.jl compatible (which is why there's the dependency on RecipesBase).

Idk about OpenStreetMap, but [OpenStreetMap2.jl](https://github.com/yeesian/OpenStreetMap2.jl) is not going to do any plotting (for now).

juliohm:
Thank you, I will stay tuned. Map tiles is a very nice package to have, I've just cloned the repo to give it a try.

yeesian:
This should now work:
```julia
using GeoJSON, Plots
# ;wget https://raw.githubusercontent.com/awoodruff/leaflet-intro/gh-pages/neighborhoods.geojson
plot(GeoJSON.parsefile("neighborhoods.geojson"))
```
which is nice, because the neighborhoods contains a mixture of Polygons and MultiPolygons, which users in Python usually have to distinguish between.

Michael Martijn and Julio can you all help review this PR one last time? Thanks so much for all the feedback. I think this PR is much better for it.

src/plotrecipes.jl
```diff
@@ -0,0 +1,105 @@
...
+end
+
+function shapecoords(geom::AbstractLineString)
+    coords = coordinates(geom)
+    first.(coords), last.(coords)
+end
+RecipesBase.@recipe f(geom::AbstractLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
+
+function shapecoords(geom::Vector{<:AbstractLineString})
+    x = Vector{Float64}[]; y = Vector{Float64}[]
```
juliohm: Should the brackets be parenthesis here or a vector of vector is intended?

yeesian: a vector of vectors

juliohm: Awesome, thanks for double checking.

visr:
multi
Small issue that can be improved on later:
If something is a multigeometry, I generally expect the separate parts to be styled the same.
Screenshot is a comparison to QGIS.

src/plotrecipes.jl
```diff
@@ -0,0 +1,105 @@
...
+    x, y
+end
+
+function shapecoords(geom::AbstractMultiLineString)
+    coords = coordinates(geom)
+    x = Vector{Float64}[first.(line) for line in coords]
+    y = Vector{Float64}[last.(line) for line in coords]
+    x, y
+end
+RecipesBase.@recipe f(geom::AbstractMultiLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
```
mkborregaard: This should be `seriestype --> :path`, `:line` sorts x values before plotting

src/plotrecipes.jl
```diff
@@ -0,0 +1,105 @@
+shapecoords(geom::AbstractPoint) = [tuple(coordinates(geom)...)]
+RecipesBase.@recipe f(geom::AbstractPoint) = (seriestype --> :scatter; legend --> :false; shapecoords(geom))
```
mkborregaard: There should also be `aspect_ratio := 1` in all recipes, which forces units on the x and y axes to be the same – otherwise the projection will be lost.

src/plotrecipes.jl
```diff
@@ -0,0 +1,105 @@
...
+        append!(x, first.(coords)); append!(y, last.(coords))
+    end
+    x, y
+end
+
+function shapecoords(geom::AbstractLineString)
+    coords = coordinates(geom)
+    first.(coords), last.(coords)
+end
+RecipesBase.@recipe f(geom::AbstractLineString) = (seriestype --> :line; legend --> :false; shapecoords(geom))
```
mkborregaard: see below on :line

src/plotrecipes.jl
```diff
@@ -0,0 +1,105 @@
...
+function shapecoords(geom::Vector{<:AbstractMultiLineString})
+    x = Vector{Float64}[]; y = Vector{Float64}[]
+    for g in geom, line in coordinates(g)
+        push!(x, first.(line)); push!(y, last.(line))
+    end
+    x, y
+end
+
+function shapecoords(geom::AbstractPolygon)
+    ring = first(coordinates(geom)) # currently doesn't plot holes
```
mkborregaard: holes are tricky.

yeesian:
Thanks Michael and Martijn, really appreciated! Your remarks should be addressed in https://github.com/JuliaGeo/GeoInterface.jl/pull/12/commits/0d43979356e6abb428ac341b203d7854096b2c77 and https://github.com/JuliaGeo/GeoInterface.jl/pull/12/commits/5c7ffd19083d7cf10a4334c1416accd7ea4d05a6 respectively.

yeesian:
I believe this should be good enough to capture many common cases for a start. Feel free to file any plotting issues in the downstream packages in this repo.

yeesian:
> This `append` seems to work fine, but I guess you meant `push` here?

Haha yes https://github.com/JuliaGeo/GeoInterface.jl/pull/12/commits/c5fad04a84040dfe8d9f87ec459f9292dd665d9a

PR #13, opened 2018-01-04
Remove unused `shapecoords()` methods

yeesian:
they were wrong (and unused anyway; since only the recipe for
`geom::Vector{<:AbstractGeometry}` is used)

Martijn, can you help me verify on

```julia
;cd ~/.julia/v0.6/Shapefile/test

using Shapefile, Plots

f(i) = plot(open("shapelib_testcases/test$i.shp") do f
    read(f, Shapefile.Handle).shapes
end, alpha=0.5)

plot([f(i) for i in 1:12]...)
```
It should give
shapefiletests

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -6,17 +6,9 @@ RecipesBase.@recipe f(geom::AbstractPoint) = (
     shapecoords(geom)
 )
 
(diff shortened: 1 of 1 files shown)


visr:
image
This is what I see. Looks ok?

yeesian:
yup!

PR #14, opened 2018-01-09
Implementation of shapecoords for AbstractGeometry

yeesian:
this is for packages implementing GeoInterface at the level of
AbstractGeometry, rather than the subtypes
AbstractPoint…AbstractMultiPolygon. They just need to define
`geotype(::AbstractGeometry)` in order for this package to identify
which geometry type it is, and return the corresponding coordinates.

We first define methods (`pointcoords()`,…,`multipolygoncoords()`) on
AbstractGeometry that returns the coordinates for different geometry
types. We can then have a `shapecoords(geom::AbstractGeometry)` that
uses `geotype(geom)` to look up the corresponding method to call

See https://github.com/yeesian/ArchGDAL.jl/commit/a5cbe25111a840c6217ac650d9b0c12216aa19d7 for an example of how it works

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -1,34 +1,22 @@
-shapecoords(geom::AbstractPoint) = [tuple(coordinates(geom)...)]
-RecipesBase.@recipe f(geom::AbstractPoint) = (
-    aspect_ratio := 1;
-    seriestype --> :scatter;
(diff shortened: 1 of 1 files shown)


visr:
Check, this is very useful. Don't see anything wrong with it. :)

Issue #17, opened 2018-10-30
MultiLineString() throws UndefVarError: line not defined


cchderrick:
Should be `MultiLineString(coordinates(lines))` instead of `MultiLineString(coordinates(line))`
https://github.com/JuliaGeo/GeoInterface.jl/blob/02bb014ff6a05b3a4d2c4dbfe3a8059ea9daf327/src/geotypes.jl#L80

visr:
Oh yes good catch, thanks! Feel free to directly create a PR if the mistake is clear like here.

visr:
Fixed by #18

PR #18, opened 2018-10-30
fix typo #17

cchderrick:


yeesian:
Thanks! I should add tests for this package at some point x)

PR #22, opened 2019-04-03
Add comparison operators for some geometry types.

### LINKED ISSUE #21: `==` not implemented for any (abstract) types
I was just surprised by
```julia
julia> [0.0, 0.0] == [0.0, 0.0]
true

julia> Point([0.0, 0.0]) == Point([0.0, 0.0])
false
```

Turns out that `==` is not implemented for any of the types in GeoInterface, so Julia applies the `===` fallback. Is this intentional?

I think a value-based comparison could be had quite easily:
```julia
julia> import Base: ==

julia> ==(x::AbstractGeometry, y::AbstractGeometry) = coordinates(x) == coordinates(y)
== (generic function with 155 methods)

julia> Point([0.0, 0.0]) == Point([0.0, 0.0])
true
```

This would be useful for me, as I try to use `Point`s as keys to `Dict`s, but I don't want to commit type piracy. I can open PR if this change is welcome.

leethargo:
Fixes #21.

Only implementing the uncontroversial part now, for `Point`s.

### DIFF
--- a/src/GeoInterface.jl
+++ b/src/GeoInterface.jl
@@ -80,6 +80,7 @@ module GeoInterface
     bbox(obj::AbstractFeatureCollection) = nothing
     crs(obj::AbstractFeatureCollection) = nothing
 
+    include("operations.jl")
     include("geotypes.jl")
     include("plotrecipes.jl")
 end
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -1,7 +1,10 @@
-import Base: ==, hash
+import Base: ==, hash, isapprox
 
 # Compare points by coordinate values.
 ==(x::P, y::P) where {P <: AbstractPoint} = coordinates(x) == coordinates(y)
 
 # Hash the coordinates for consistency.
 hash(x::P) where {P <: AbstractPoint} = hash(coordinates(x))
+
+# Compare points approximately by coordinate values.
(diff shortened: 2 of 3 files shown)


leethargo:
Personally, I would generalize this for `AbstractGeometry` everywhere (rather than just `AbstractPoint`), but my personal use case is covered, so I'm happy either way.

yeesian:
Thanks for the PR!

> Personally, I would generalize this for AbstractGeometry everywhere (rather than just AbstractPoint), but my personal use case is covered, so I'm happy either way.

I just want to add that I'm in personal agreement with you here (since it has a precedent in [shapely](https://github.com/Toblerity/Shapely/issues/535#issuecomment-339612312)).

But discussion has been happening in places like https://github.com/JuliaGeometry/GeometryTypes.jl/pull/166 and in the geo channel on Slack  (some of them summarised [here](https://github.com/JuliaGeometry/meta/wiki/Multi-Geometries) and [here](https://github.com/JuliaGeometry/meta/wiki/Metadata-for-Geometries)) that gives me pause on whether the `coordinates()` method is the appropriate interface for people's needs (and hence whether it is the appropriate thing to define a measure of equality on), so I'd prefer to hold off on large changesets in this package for now.

leethargo:
Fair enough. The discussions are long and over my head. Maybe there will be some nice convergence by JuliaCon :smiley:

Issue #23, opened 2019-06-05
Why dispatch on concrete types?


rafaqz:
I'm wondering why `Vector` is used everywhere for dispatch, instead of `AbstractVector`? and `Float64` instead of `Number`?

visr:
I don't think there are hard reasons for this. It probably was easiest to start off with concrete types to hash out a first design, but it would be good to work towards making this more generic.

yeesian:
https://github.com/JuliaGeo/GeoInterface.jl/issues/20#issuecomment-458564370 was what I was thinking when I first wrote this package. That thread is also the latest discussion on it that I know of. I'm currently bound up in graduation plans, but PRs are definitely welcome!

rafaqz:
Ok that makes sense. I'll look at doing a refactor. 

It also seems we could cut half the code in this package without losing functionality. It would be more inviting as an api if it was simpler.

rafaqz:
I've just done most of a massive refactor. But I just realised there aren't enough tests to verify a pull request, and I've definately made a mistake or two.

Would running the tests for Shapefile.jl, ArchGDAL and GeoJSON be adequate?

visr:
Yeah testing those packages would be good, and I'd also add LibGEOS.

rafaqz:
Some comments on generalising GeoInterface:

`Position` not being `<: AbstractPosition` is a barrier to abstraction, as we can't add methods to the constructor for Position to handle bad inputs

So letting `MultiPolygon` and similar hold other kinds of positions (say StaticArrays) isn't so easy.  It works, but generalising everything loses feature you're currently getting for free like converting from arrays of Any (returned by JSON.jl) to arrays of Float64 with `Position[]`. 

It's also it harder to simplify dispatch because we always need to use `Union{Position, AbstractPosition}` which leads to ambiguities, that `Position <: AbstractPosition` would resolve.

visr:
Yeah I can understand that it's hard to generalize this completely without breaking something. That's why it is good to discuss and try out approaches. I think we all agree that `const Position = Vector{Float64}` simple as it is, does not have the future :). We also seem to agree that `Vector{Float64}` should definitely be accepted as a position, just like `SVector{3, Float32}`.

[...]

PR #25, opened 2019-08-03
Refactor

rafaqz:
This isn't ready to pull yet, more as a placeholder/reminder that this refactor has been done. 

Merging it is blocked by the GeoJSON test suite. It doesn't handle its own `Any` typed positions and relies on the `Vector{Float64}` type conversion that previously happened in GeoInterface.jl, but really shouldn't. It shouldn't be too much hassle to fix that, I just haven't gotten around to it.

It would also be good if we ran the test suites for dependent packages in the CI for this package.

Edit: its probably worth just waiting and using refactoring to use GeometryTypes now that I've looked into it, I mostly refactored this just to learn the package so its not really wasted effort.

### DIFF
--- a/src/GeoInterface.jl
+++ b/src/GeoInterface.jl
@@ -2,85 +2,28 @@ __precompile__()
 
 module GeoInterface
 
-    using RecipesBase
+using RecipesBase
 
-    export  AbstractPosition, Position,
-            AbstractGeometry, AbstractGeometryCollection, GeometryCollection,
-            AbstractPoint, Point,
-            AbstractMultiPoint, MultiPoint,
-            AbstractLineString, LineString,
-            AbstractMultiLineString, MultiLineString,
-            AbstractPolygon, Polygon,
-            AbstractMultiPolygon, MultiPolygon,
-            AbstractFeature, Feature,
-            AbstractFeatureCollection, FeatureCollection,
+export  AbstractPosition, Position,
+        AbstractGeometry, AbstractGeometryCollection, GeometryCollection,
+        AbstractPoint, Point,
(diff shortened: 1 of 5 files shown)


yeesian:
> Edit: its probably worth just waiting and using refactoring to use GeometryTypes

Feel free to ping me and I'll be happy to review this PR when you are done. The GeometryTypes effort has a long history and requires a level of coordination that I don't think should stall other efforts at smaller improvements in the meantime.

rafaqz:
Ok that would be great. Maybe getting things a little more organised will help the transition as well. But I'm mostly focussed on https://github.com/rafaqz/GeoDataBase.jl (kind of xarray for julia) so this will probably sit here for a while.

yeesian:
> But I'm mostly focussed on https://github.com/rafaqz/GeoDataBase.jl (kind of xarray for julia) so this will probably sit here for a while.

You might be interested in https://github.com/evetion/GeoArrays.jl too

visr:
> Merging it is blocked by the GeoJSON test suite.

GeoJSON is I believe the only package relying on the concrete types rather than the interface of GeoInterface. Perhaps it shouldn't and this package should only aim to be an interface for the diverse set of types that want to adhere to this interface.

We talked a bit about this at https://github.com/JuliaGeo/GeoInterface.jl/issues/23#issuecomment-500501747 and your reply there:

> I also agree on moving some responsibility to GeoJSON. Specifically converting Vector{Any} to Vector{Float64}

To keep track of this, and for discussion, I just opened https://github.com/JuliaGeo/GeoJSON.jl/issues/20 and https://github.com/JuliaGeo/GeoJSON.jl/issues/21.

PR #27, opened 2019-12-04
add a Project.toml

visr:
Drops julia 0.7 support and updates CI as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
+  - 1.0
+  - 1
   - nightly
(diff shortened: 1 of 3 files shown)


yeesian:
LGTM

.travis.yml
```diff
@@ -3,14 +3,11 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
+  - 1.0
+  - 1
```
yeesian: what's the distinction between 1.0 and 1?

visr: 1 in this case means the latest point release, so 1.3 right now.

PR #28, opened 2020-01-17
use warn macro

jaakkor2:
FWIW, in Julia v0.7 `warn` was deprecated, changed to `@warn`.

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -121,7 +121,7 @@ RecipesBase.@recipe function f(geom::AbstractGeometry)
     elseif gtype == :Polygon || gtype == :MultiPolygon
         seriestype := :shape
     else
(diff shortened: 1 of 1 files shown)


visr:
Ah thanks for the fix. This `warn` now errors on Julia 1.0.

I guess this comes from trying to plot a `GeometryCollection`? That's the only subtype not covered by the recipe it seems. You should be able to plot all the elements of the `GeometryCollection` though. Maybe that's actually not to so hard to add to the recipe.

jaakkor2:
Plotting a `GeometryCollection` works fine.  [ I was trying to figure some things out with `GeoInterface.shapecoords(geomcollection)` when I noticed the `warn` error. ]

visr:
Ok yeah now I see that `GeometryCollection` is defined below the others. Merging and tagging this.

PR #38, opened 2021-04-12
allow setting of aspect_ratio

rush42:
I think there are several use cases where one wants to specify an other aspect ratio than 1, e.g. when plotting geometries in spherical coordinates.

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -69,49 +69,49 @@ function shapecoords(geom::AbstractGeometry)
 end
 
 RecipesBase.@recipe f(geom::AbstractPoint) = (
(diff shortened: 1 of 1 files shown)


visr:
Thanks for the contribution! I'm not too familiar with the recipe syntax, but is it correct that this keeps the default at 1, but allows it to be overridden by the user. Did you test this for yourself already?

rush42:
Yes, this keeps aspect_ratio set to 1 unless the user specifies it himself. You can see it in the examples of RecipesBase -> A real example (https://github.com/JuliaPlots/RecipesBase.jl). I tested it and it works as expected.

PR #43, opened 2022-01-07
only set aspect_ratio on nonmutating calls of plot()

rush42:
This PR addresses the following issue:

Assume the following code example:
```julia
using GeoInterface, Plots

plot(1:3,1:3, aspect_ratio=2, legend=true)
ls = LineString([[0.,0.],[1.,0.],[1.,1.],[0.,1.], [0.,0.]])
plot!(ls)
```

The mutating call `plot!(ls)` resets the previous set `aspect_ratio` and `legend` arguments to the value defined in the recipe of this package.

This PR changes the above described behavior in the following way:

for `aspect_ratio`:
only set the `aspect_ratio` for the first plot series and not for mutating calls of `plot!()`

for `legend`:

replace `legend --> false` with `label --> :none`

this ensures that there will be no `legend` as long as you only add subtypes of  `AbstractGeometries` without an explicit `label`

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -69,63 +69,63 @@ function shapecoords(geom::AbstractGeometry)
 end
 
 RecipesBase.@recipe function f(geom::AbstractPoint)
+    label --> :none
     if plotattributes[:plot_object].n == 0 
         aspect_ratio --> 1
-        legend --> :false
     end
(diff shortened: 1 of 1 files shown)


visr:
Not too familiar with the plotting recipes, but I tried it out with your example and it looks like a clear improvement. Thank you!

PR #47, opened 2022-05-01
add getpolygons and getrings functions

rafaqz:
This PR adds `getpolygons` and `getrings` functions to allow efficient access to all poylgons and rings in cases like Shapefile.jl where exterior and holes are not clearly separated in the spec. Packages like Rasters.jl don't actually care if they are separated so loading them all directly is much more efficient. 

The fallback is to iterate over all polygons and rings.

### DIFF
--- a/src/defaults.jl
+++ b/src/defaults.jl
@@ -50,7 +50,7 @@ getpolygon(p::AbstractMultiPolygonTrait, geom) = getgeom(p, geom)
 getpolygon(p::AbstractMultiPolygonTrait, geom, i) = getgeom(p, geom, i)
 getring(g::AbstractMultiPolygonTrait, geom) = (r for r in getring(p) for p in getpolygon(geom))
 getpoint(g::AbstractMultiPolygonTrait, geom) = (p for p in getpoint(r) for r in getring(geom))
-nring(p::AbstractPolygonTrait, geom) = sum(nring(p) for p in getpolygon(p))
+nring(p::AbstractMultiPolygonTrait, geom) = sum(nring(p) for p in getpolygon(p))
 
 ## Surface
 npatch(p::AbstractPolyHedralSurfaceTrait, geom)::Integer = ngeom(p, geom)
--- a/src/defaults.jl
+++ b/src/defaults.jl
@@ -30,27 +30,31 @@ getpoint(c::AbstractCurveTrait, geom, i) = getgeom(c, geom, i)
 startpoint(c::AbstractCurveTrait, geom) = getpoint(c, geom, 1)
 endpoint(c::AbstractCurveTrait, geom) = getpoint(c, geom, length(geom))
 
-## MultiLineString
-nlinestring(p::AbstractMultiLineStringTrait, geom) = ngeom(p, geom)
-getlinestring(p::AbstractMultiLineStringTrait, geom) = getgeom(p, geom)
-getlinestring(p::AbstractMultiLineStringTrait, geom, i) = getgeom(p, geom, i)
-getpoint(g::AbstractMultiLineStringTrait, geom) = (p for p in getpoint(p) for g in getlinestring(geom))
-
 ## Polygons
 nring(p::AbstractPolygonTrait, geom) = ngeom(p, geom)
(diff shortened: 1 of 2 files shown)


evetion:
In the current documentation I had written:
> Where the getgeom could be an iterator (without the i) as well.

So that would be your functions, but without the `s` at the end. `getgeom(geom)` would give you an iterator and the first geom, which naturally falls back to the `getgeom(geom, i)` methods?

rafaqz:
Ah ok I messed that. Sounds good to me. We can do the same for `getring`.

evetion:
I've added iterators in the main pr https://github.com/JuliaGeo/GeoInterface.jl/pull/33/commits/1ba01f5c6f824ae1fe7509f17c9fd0016012a313. Only your `getrings(g::AbstractMultiPolygonTrait, geom) = (r for r in getrings(p) for p in getpolygons(geom))` that iterates over subgeometries as well is missing, as I think its better suited at ShapeFile itself?

rafaqz:
> I think its better suited at ShapeFile itself?

The idea is that Rasters.jl can call it without depending on Shapefile.jl, and `getring` still works for other types.

rafaqz:
Ok I've rebased and simplified this.

src/defaults.jl
```diff
@@ -47,6 +47,8 @@ getlinestring(p::AbstractMultiLineStringTrait, geom, i) = getgeom(p, geom, i)
 npolygon(p::AbstractMultiPolygonTrait, geom) = ngeom(p, geom)
 getpolygon(p::AbstractMultiPolygonTrait, geom) = getgeom(p, geom)
 getpolygon(p::AbstractMultiPolygonTrait, geom, i) = getgeom(p, geom, i)
+getring(g::AbstractMultiPolygonTrait, geom) = (r for r in getrings(p) for p in getpolygons(geom))
+nring(p::AbstractPolygonTrait, geom) = sum(nring(p) for p in getpolygons(p))
```
evetion, line 51:
```suggestion
nring(p::AbstractMultiPolygonTrait, geom) = sum(nring(p) for p in getpolygons(p))
```

src/interface.jl
```diff
@@ -313,17 +318,31 @@ Note that this is only valid for [`AbstractMultiPolygon`](@ref)s.
...
 
 """
-    getpolygon(geom, i::Integer) -> AbstractCurve
+    getpolygon(geom, [i::Integer]) -> AbstractCurve
 
 Returns the `i`th polygon for the given `geom`.
+Where no `i` index is passed, an iterator over all polygons is returned.
 Note that this is only valid for [`AbstractMultiPolygon`](@ref)s.
 """
+getpolygon(geom) = getpolygon(geomtype(geom), geom)
```
evetion, line 327: Duplicate of r332.

src/interface.jl
```diff
@@ -313,17 +318,31 @@ Note that this is only valid for [`AbstractMultiPolygon`](@ref)s.
...
 getpolygon(geom, i::Integer) = getpolygon(geomtype(geom), geom, i)
 """
     getpolygon(geom) -> Iterator
 """
 getpolygon(geom) = getpolygon(geomtype(geom), geom)
 
+"""
```
evetion, line 334: Complete duplication r200 onwards.

evetion (changes requested):
The requested functionality is like adding flattening, but not completely, like coordinates, but over a single level. Should we make this consistent and also add getpoint/npoint for MultiLineStrings?

src/interface.jl
rafaqz: ahh ok rebase didn't work so well

rafaqz:
> Should we make this consistent and also add getpoint/npoint for MultiLineStrings?

Sure, then do we have to add it `getpoint` and `npoint` for `Polygon` and `MultiPolygon` too? (that works for me).

rafaqz:
Ok added `getpoint` everywhere it seems to makes sense, and standardised the iterator function docs.

evetion:
LGTM. If anything, I doubt about the `getpoint` on the `MultiPolygon` one, since that's three levels. But it still feels like a useful thing.

PR #50, opened 2022-05-15
export the traits

visr:
This is onto the branch of #49, but separate because it might warrant some discussion.

One of the advantages of adding *Trait to all trait types, is that they wont clash with other packages. The disadvantage is that some of them get even longer, e.g. `GeoInterface.AbstractGeometryCollectionTrait`. But we can alleviate this by exporting all the traits. What do you think?

For my own reference, replacements made with

```
GeoInterface.(\w+)Trait
$1Trait
```


rafaqz:
Yes having Trait at the end means they're really unlikely to clash, so it's probably a good idea

PR #57, opened 2022-05-16
Add implementations for Base types.

### LINKED ISSUE #54: Fallback point implementations
Currently nothing from Base is a geometry by default, in the way that in Tables.jl a Vector of NamedTuple or NamedTuple of Vector is a table. Would it made sense to support some? It might be nice if e.g. a SVector or NTuple would be considered a point.

It looks like it would only take:

```julia
GeoInterface.isgeometry(::AbstractVector{<:Real}) = true
GeoInterface.geomtype(::AbstractVector{<:Real}) = PointTrait()
GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = length(geom)
GeoInterface.getcoord(::PointTrait, geom::AbstractVector{<:Real}, i) = getindex(geom, i)

GeoInterface.isgeometry(::NTuple{N, <:Real}) where {N} = true
GeoInterface.geomtype(::NTuple{N, <:Real}) where {N} = PointTrait()
GeoInterface.ncoord(::PointTrait, geom::NTuple{N, <:Real}) where {N} = N
GeoInterface.getcoord(::PointTrait, geom::NTuple{N, <:Real}, i) where {N} = getindex(geom, i)
```

Perhaps we could even do this for AbstractVector of AbstractVector, e.g. the old GeoInterface concrete GeoJSON based types would still fall under the new interface.

https://github.com/JuliaGeo/Proj4.jl/pull/64

evetion:
Fixes #54

### DIFF
--- a/docs/src/guides/defaults.md
+++ b/docs/src/guides/defaults.md
@@ -51,11 +51,11 @@ GeoInterface is implemented automatically for `NTuple`s, `NamedTuple`s and `Vect
 
 ```julia
 a = [1,2,3]
-x(a) == 1
+GeoInterface.x(a) == 1
 
 b = (1,2,3)
-x(b) == 2
+GeoInterface.y(b) == 2
 
-c = (;X=1, Y=2)
-x(c) == 1
+c = (;X=1, Y=2, Z=3)
+GeoInterface.z(c) == 3
 ```
--- a/docs/src/guides/defaults.md
+++ b/docs/src/guides/defaults.md
@@ -45,3 +45,17 @@ npoint(::QuadTrait, geom) = 4
 npoint(::PentagonTrait, geom) = 5
 npoint(::HexagonTrait, geom) = 6
 ```
+
+# Implementations
+GeoInterface is implemented automatically for `NTuple`s, `NamedTuple`s and `Vector`s to behave as Points.
+
+```julia
+a = [1,2,3]
+x(a) == 1
+
+b = (1,2,3)
+x(b) == 2
+
+c = (;X=1, Y=2)
+x(c) == 1
+```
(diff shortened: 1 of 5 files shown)


src/base.jl
```diff
@@ -0,0 +1,19 @@
...
+GeoInterface.geomtrait(::AbstractVector{<:Real}) = PointTrait()
+GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom)
+GeoInterface.getcoord(::PointTrait, geom::AbstractVector{<:Real}, i) = getindex(geom, i)
+
+GeoInterface.isgeometry(::Type{<:NTuple{N,<:Real}}) where {N} = true
+GeoInterface.geomtrait(::NTuple{N,<:Real}) where {N} = PointTrait()
+GeoInterface.ncoord(::PointTrait, geom::NTuple{N,<:Real}) where {N} = N
+GeoInterface.getcoord(::PointTrait, geom::NTuple{N,<:Real}, i) where {N} = getindex(geom, i)
+
+for i in 2:4
```
rafaqz, line 13: This assumes default point order. What about (:Y, :X) points?

docs/src/guides/defaults.md
```diff
@@ -45,3 +45,17 @@ npoint(::QuadTrait, geom) = 4
 npoint(::PentagonTrait, geom) = 5
 npoint(::HexagonTrait, geom) = 6
 ```
+
+# Implementations
+GeoInterface is implemented automatically for `NTuple`s, `NamedTuple`s and `Vector`s to behave as Points.
```
visr, line 50: Perhaps good to be precise; Vector should be AbstractVector, and eltype for all should be a Real.

rafaqz:
LGTM

PR #61, opened 2022-06-01
`Feature` and `isfeaturecollection`

rafaqz:
This PR rounds out features a little more, adding `isfeaturecollection` and a wrapper object `Feature`.

An `AbstractVector{Feature}` is then defined as the basic default feature collection.

Also updates the tests to better use the new defaults and `testfeature` and `testfeaturecollection`.


rafaqz:
Ok yes it is a little confusing using `geomtrait`. It's wont' make any difference for performance, it just lets us standardise method dispatch instead of always needing a block like this everywhere: 

```julia
if GeoInterface.isfeature(x) 
    ... 
elseif GeoInterface.isfeaturecollection(x) 
    ...
elseif GeoInterface.isgeometry(x) 
    ... 
end
```

evetion:
GeomTrait and FeatureTrait and FeatureCollectionTrait as options for a trait method?

rafaqz:
Yep. I added the `trait` method here to demonstrate as it's only a few lines. Works well to apply the same methods on both features and geometries.

evetion:
Looks good! Is the `NamedTupleFeature` something you use yourself?

rafaqz:
Yeah Im experimenting with what to return from `extract` in Rasters.jl. It should be some kind of feature.

Also we need something to run the tests on here, the reason its in this PR and not separate.

rafaqz:
Lets not register this yet Im not 100% sure about it

PR #64, opened 2022-06-20
Features fixes

rafaqz:
This changes a few things from the last PR #61 

Features are now defined for any `NamedTuple` that has a `geometry` field, and all other fields are `properties`.

An `AbstractTrait` type is also added to unify all the traits.

Some of these methods like `geometry` and `nfeature` are returning `nothing` where they could instead throw a `MethodError` like another object without the method defined. I'm not sure which is best here.


evetion:
That would be inline with the default interface for geometry and properties, they return `nothing` by default.

evetion (approved):

PR #67, opened 2022-08-15
Added fallback method for `extent`.

evetion:
While slow, it's very useful for packages that don't implement it itself, as we use `extent` in packages such as https://github.com/evetion/GeoAcceleratedArrays.jl/pull/3

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,13 +1,13 @@
 name = "GeoInterface"
 uuid = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
 authors = ["JuliaGeo and contributors"]
-version = "1.0.1"
+version = "1.0.2"
 
 [deps]
 Extents = "411431e0-e8b7-467b-b5e0-f676ba4f2910"
 
 [compat]
-Extents = "0.1"
+Extents = "0.1.1"
 julia = "1"
 
 [extras]
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -411,13 +411,12 @@ In SF this is defined as `SRID`.
 crs(geom) = crs(geomtrait(geom), geom)
 
 """
-    extent(geom) -> T <: Extents.Extent
+    extent(obj) -> T <: Extents.Extent
 
-Retrieve the extent (bounding box) for given geom.
+Retrieve the extent (bounding box) for given geom or feature.
 In SF this is defined as `envelope`.
 """
-extent(geom) = extent(geomtrait(geom), geom)
-extent(trait, geom) = nothing
+extent(obj) = extent(trait(obj), obj)
 
 """
     bbox(geom) -> T <: Extents.Extent
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -411,12 +411,19 @@ In SF this is defined as `SRID`.
 crs(geom) = crs(geomtrait(geom), geom)
 
 """
-    extent(obj) -> T <: Extents.Extent
+    extent(obj, fallback=true) -> T <: Extents.Extent
 
 Retrieve the extent (bounding box) for given geom or feature.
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -1,13 +1,13 @@
 name = "GeoInterface"
 uuid = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
 authors = ["JuliaGeo and contributors"]
-version = "1.0.1"
+version = "1.0.2"
```
rafaqz, line 4: Best not to bump versions in PRs because we always forget and bump again later, e.g. for the last ArchGDAL version.

rafaqz (commented):
LGTM 

The one thing missing with this is a way for other packages to know if the extent exists already or has to be generated, like `GeoInterface.hasknownextent`. 

As an example use case for this, in Rasters.jl I could could calculate the extent as the geometry is loaded to avoid doing it twice.

evetion:
> The one thing missing with this is a way for other packages to know if the extent exists already or has to be generated, like `GeoInterface.hasknownextent`.
> 
> As an example use case for this, in Rasters.jl I could could calculate the extent as the geometry is loaded to avoid doing it twice.

But if you calculate it beforehand, you could just dispatch on it? In essence that would be (like) a prepared geometry. Of course, that would require a type being aware of it, but that feels more inline with what we've been doing thus far?

rafaqz:
I mean, for example shapefile geoms have known extents, geojson geoms may or may not.  With shapefiles you can always use that knowledge to decrease load time by ignoring a lot of things before you do any processing at all.
Rasters.jl will just return an empty array if there are no extents covering the target dimensions in `rasterize`.

With GeoJSON currently you also know `extent` has no overhead because it will just return `nothing` if there is no bbox and you can decide if calculating it makes sense.

With this PR we lose that information, and can not trust `extent` to be a free optimisation anywhere.

I just wanna know that somehow or other to avoid double reads of the data.

evetion:
Hmmm, both arguments make sense to me. Yes, I'd like to know `extent` is performance stable(ish). But I also want to get an extent when I call extent, not `nothing`.

`extent(geom; fallback=false)` by default, when `fallback=true`, we use the method in this PR, which only triggers when the geom doesn't implement it itself? Fallback could also be named calculate.

`hasknownextent` requires yet another method to be implemented and is directly out of sync with all the packages that already define `extent` but didn't update yet.

rafaqz:
A keyword could work. The default could be `true` so getting an `Extent` is the normal behavior.

src/interface.jl
```diff
@@ -411,12 +411,19 @@ In SF this is defined as `SRID`.
 crs(geom) = crs(geomtrait(geom), geom)
 
 """
-    extent(obj) -> T <: Extents.Extent
+    extent(obj, fallback=true) -> T <: Extents.Extent
```
rafaqz, line 414:
```suggestion
    extent(obj; fallback=true) -> Extents.Extent
```

rafaqz (approved):
LGTM

visr:
I'm a bit confused on the `extent` function. If I'm not mistaken both Extents and GeoInterface have a separate `extent` function. Shouldn't GeoInterface add methods to the `Extents.extent` function? Or is that deliberately not done?

evetion:
GI.extent -> Extents.Extent(). So the last (with an uppercase E) is just the type constructor.

visr:
Yeah that part is clear. It's just confusing to me that both packages define `extent`, but also other functions like `intersect`.

rafaqz:
Yeah it's a bit weird. 

But if we used `Extents.extent` we couldnt add this method without type piracy, so I guess it was a good decision.

Issue #70, opened 2022-10-09
Makie support?


jw3126:
There is GeoInterfaceRecipes which gives Plots.jl support for anything that implements GeoInterface.
Is there something similar for Makie support?
I did not find anything and put together a small [package](https://github.com/jw3126/GeoInterfaceMakie.jl) myself. 
Just want to double check I am not reinventing wheels here before registering that package.

visr:
That's interesting, thanks for sharing. Are you aware of [GeoMakie](https://github.com/MakieOrg/GeoMakie.jl)? Makie itself uses [GeometryBasics](https://github.com/JuliaGeometry/GeometryBasics.jl/) geometry types, and since GeometryBasics implements the GeoInterface as well, this should allow GeoMakie to accept other geometry types as well.

jw3126:
What I am interested in doing is plotting LibGEOS stuff with Makie. I looked briefly at GeoMakie before it looks like a powerful package. However my impression was that it is geared towards plotting geographical data and does not really help with plotting say a LibGEOS.Polygon?

rafaqz:
That's great. This actually came up with @SimonDanisch the other day.

If you give me a few days I can go over the code if that would help.

And do you want to put it in JuliaGeo?

jw3126:
> If you give me a few days I can go over the code if that would help.

Awesome @rafaqz I would appreciate that.

> And do you want to put it in JuliaGeo?

Sure! Do you think it makes sense to put it here in the repo as a subpackage analogous to GeoInterfaceRecipes?

rafaqz:
Yeah it can go in here as well, that's even simpler.

visr:
> my impression was that it is geared towards plotting geographical data and does not really help with plotting say a LibGEOS.Polygon

I'd think the latter would also be a usecase for GeoMakie, although most examples are world maps. If you prefer to have a separate GeoInterfaceMakie that does not deal with projections and such, we could make GeoMakie depend on GeoInterfaceMakie. Only reason I'm slightly hesistant is that I don't see much advantage in splitting the packages other than avoiding perhaps a PROJ dependency.

rafaqz:
I guess we need to think about what default plots should be.

I feel that simple 2d plots without projections to be the default? and just using `plot(geom)` rather than defining a `GeoAxis` manually?

[...]

PR #71, opened 2022-10-11
add GeoInterfaceMakie

jw3126:


GeoInterfaceMakie/src/GeoInterfaceMakie.jl
```diff
@@ -0,0 +1,101 @@
...
+    Makie.Poly
+end
+function pttype(geom)
+    if GI.is3d(geom)
+        GB.Point3f
+    else
+        GB.Point2f
+    end
+end
+function points(geom)::Union{Vector{GB.Point2f}, Vector{GB.Point3f}}
```
jw3126, line 30: TODO: 
https://github.com/JuliaGeo/GeoInterface.jl/issues/70#issuecomment-1272649198

jw3126:
TODO I guess I should add CI for this subpackage.

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
jw3126: @rafaqz is this what you had in mind?

.github/workflows/CIGeoInterfaceMakie.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: julia-actions/cache@v1
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: Run julia tests
+        shell: julia --project=GeoInterfaceMakie
+        run: |
```
jw3126, line 38: This looks a bit hacky. Is this the right way to do CI for subpackages?

rafaqz: Hah I don't even know, I just realised we probably aren't testing the Plots.jl interface eithere

rafaqz: Oh no its actually here:
https://github.com/JuliaGeo/GeoInterface.jl/blob/main/.github/workflows/recipes.yml

Probably copy that

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
rafaqz: Sorry I should have been more specific. 

You can call `GeoInterface.getpoint(obj)` and get an iterator over all the points in nested geometries. This has optimizations for e.g. shapefile, where it will just give you the vectors of points that are already iterators.

.github/workflows/CIGeoInterfaceMakie.yml
jw3126: It seems that file never ran, hopefully I fixed that.

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
```diff
@@ -0,0 +1,116 @@
...
+function _convert_arguments(::MC.PointBased, geom)::Tuple
+    pts = points(geom)
+    (pts,)
+end
+
+function expr_enable(Geom)
+    quote
+        # import GeoInterfaceMakie.MakieCore
+        # import GeoInterfaceMakie.Makie
+        function $Makie.plottype(geom::$Geom)
```
rafaqz, line 79: `plottype` is in MakieCore, can we remove the Makie.jl dependency? We can only really use MakieCore in JuliaGeo packages.

jw3126: You are right about this one, however sadly Poly is in Makie and not in MakieCore.

rafaqz: Damn. But we can't have everything depend on Makie.jl. 

I'll make an issue over there to get it moved

jw3126: I already made a PR there, no idea if it gets merged or what the obstacles are.

rafaqz: Oh thanks. 

I'm not sure why anything returned by `plottype` is not in MakieCore, it kind of defeats the purpose of having a recipe system.

jw3126:
@rafaqz poly is now available in MakieCore and tests pass here without depending on Makie!

rafaqz (approved):
LGTM

jw3126:
Thanks for merging! Can you register GeoInterfaceMakie?

rafaqz:
Sure! Kind of annoying for you writing a package and you cant administer it. Do you want to be a moderator/admin here?

jw3126:
Yes would be great if I could do more things myself here!

PR #72, opened 2022-10-12
add Plots.jl recipes for features and feature collections

rafaqz:
@jw3126 you may want to add this for Makie as well


evetion (approved):

evetion:
LGTM. But it does remind me that we need document the usage of these recipes.

PR #74, opened 2022-10-15
run jet and fix bugs

rafaqz:
JET found a bunch of places we were not checking for zero/one length geoms. After fixing it still complains about some as it doesn't seem to see the runtime check on `npoint` means the iterator will never return `nothing`.

```julia
# Run from the GeoInteface environment
julia> JET.report_package(GeoInterface)
...
═════ 3 possible errors found ═════
┌ @ /home/raf/.julia/dev/GeoInterface/src/interface.jl:125 GeoInterface.issimple(geomtrait(geom), geom)
│┌ @ /home/raf/.julia/dev/GeoInterface/src/fallbacks.jl:90 GeoInterface.allunique(Base.Generator(#23, 1 GeoInterface.:(:) GeoInterface.npoint(t, geom) GeoInterface.:- 1))
││┌ @ set.jl:408 Base.indexed_iterate(x, 1)
│││┌ @ tuple.jl:91 x = iterate(I)
││││ no matching method found `iterate(::Nothing)`: x = iterate(I::Nothing)::Union{}
│││└───────────────
││┌ @ set.jl:408 Base.indexed_iterate(x, 2, _6)
│││┌ @ tuple.jl:96 x = iterate(I, state)
││││ no matching method found `iterate(::Nothing, ::Int64)`: x = iterate(I::Nothing, state::Int64)::Union{}
│││└───────────────
┌ @ /home/raf/.julia/dev/GeoInterface/src/fallbacks.jl:122 GeoInterface.error(string("Conversion is enabled for type ", T, ", but not implemented. Please report this issue t
o the package maintainer."))
│┌ @ error.jl:35 error(::String)
││ may throw: Base.throw(Base.ErrorException(s::String)::ErrorException)::Union{}
│└───────────────

```

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -85,7 +85,12 @@ npoint(::HexagonTrait, geom) = 6
 nring(::HexagonTrait, geom) = 1
 
 # TODO Only simple if it's also not intersecting itself, except for its endpoints
-issimple(t::AbstractCurveTrait, geom) = allunique((getpoint(t, geom, i) for i in 1:npoint(geom)-1)) && allunique((getpoint(t, geom, i) for i in 2:npoint(t, geom)))
+function issimple(t::AbstractCurveTrait, geom)
+    n = npoint(t, geom)
+    n > 1 || return true
+    allunique((getpoint(t, geom, i) for i in 1:n-1)) && 
+    allunique((getpoint(t, geom, i) for i in 2:n))
+end
 isclosed(t::AbstractCurveTrait, geom) = getpoint(t, geom, 1) == getpoint(t, geom, npoint(t, geom))
 isring(t::AbstractCurveTrait, geom) = issimple(t, geom) && isclosed(t, geom)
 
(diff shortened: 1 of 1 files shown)


yeesian (changes requested):
Thank you! Can we add a unit test such that the current code will fail and the new ode will pass?

rafaqz:
Sure. (I made 4 of these PR at once analysing Rasters.jl  so writing tests was not in the time budget)

rafaqz:
Additionally, testing this is not possible with the GeoInterface test types as they have set lengths. I think they need a rethink.

yeesian (approved):
Resolving my earlier comment about having tests in lieu of https://github.com/JuliaGeo/GeoInterface.jl/pull/75#issuecomment-1304783752, thank you for the discussion!

evetion:
I'm on the fence here, is this gonna change anything, like performance or inference? Both functions work as is, and will return the same answer (like the Any[] for coordinates).

rafaqz:
It will make performance worse unfortunately. 

I haven't merge because this is really the "what to do with zero length geometries" problem and its not even clear in most specs. So maybe we wait a little.

rafaqz:
Closing as this is really the "what to do with zero length geometries" problem

PR #78, opened 2022-11-20
Add wrapper types for all geometries

rafaqz:
This PR adds wrapper types for all geometries.

The idea is these will work with wrapping
1. Any other GeoInterface.jl compatible objects, as-is.
2. AbstractArrays of e.g. `PointTrait` objects for `LineString` or `MultiPoint`.

Basically you should be able to hack together any geometries you want using any package components and they will work.

It's not quite there yet.


rafaqz:
@evetion if you want to have a look at this it would be good to get it merged soon so we can use it in tests.

rafaqz:
This is ready to go

rafaqz:
@evetion @visr can I get a review for this? It's needed to finish the LibGEOS convert PR.

src/fallbacks.jl
```diff
@@ -22,6 +22,9 @@ isempty(::AbstractGeometryTrait, geom) = false
 ngeom(::AbstractPointTrait, geom) = 0
 getgeom(::AbstractPointTrait, geom) = nothing
 getgeom(::AbstractPointTrait, geom, i) = nothing
+getpoint(::AbstractPointTrait, geom, i) = geom # Helps generic `gepoint` work with mixed types
```
visr, line 25:
```suggestion
getpoint(::AbstractPointTrait, geom, i) = geom # Helps generic `getpoint` work with mixed types
```

src/fallbacks.jl
```diff
@@ -39,7 +42,9 @@ nhole(t::AbstractPolygonTrait, geom) = nring(t, geom) - 1
 gethole(t::AbstractPolygonTrait, geom) = (getgeom(t, geom, i) for i in 2:ngeom(t, geom))
 gethole(t::AbstractPolygonTrait, geom, i) = getring(t, geom, i + 1)
 npoint(t::AbstractPolygonTrait, geom) = sum(npoint(p) for p in getring(t, geom))
-getpoint(t::AbstractPolygonTrait, geom) = flatten((p for p in getpoint(r)) for r in getring(t, geom))
+getpoint(t::PolygonTrait, geom) = flatten((p for p in getpoint(r)) for r in getring(t, geom))
+getpoint(t::AbstractPolygonTrait, geom) = getgeom(t, geom)
+getpoint(t::AbstractPolygonTrait, geom, i) = getgeom(t, geom, i)
```
visr, line 47: Is this correct? Won't this return rings? These definitions are the same as `getring` right above.

src/wrappers.jl
```diff
@@ -0,0 +1,241 @@
+
+export Point, EmptyPoint, LineString, Polygon, Triangle, MultiPoint, MultiCurve, MultiPolygon, TIN, Collection, Feature, FeatureCollection
+
+# Implement interface
+
+"""
+    WrapperGeometry
+
+Provides geometry wrappers that accept any GeoInterface compatible
+objects. These can be usefull for building custom geometries, or in tests.
```
visr, line 10:
```suggestion
objects. These can be useful for building custom geometries, or in tests.
```

src/wrappers.jl
```diff
@@ -0,0 +1,241 @@
+
+export Point, EmptyPoint, LineString, Polygon, Triangle, MultiPoint, MultiCurve, MultiPolygon, TIN, Collection, Feature, FeatureCollection
+
+# Implement interface
+
+"""
+    WrapperGeometry
```
visr, line 7:
```suggestion
    abstract type WrapperGeometry{Z,M,T}
```

Would be good to add to the docstring that Z and M are boolean type parameters that indicate if Z and M dimensions are present, and T is the geometry type.

visr:
> should the types go in a module they can be easily imported, instead of all exported

My first thought would be to not export them as they will often clash with other packages. Putting them in a module works, perhaps combined with making them available unexported in the GeoInterface namespace as well, such that `GI.Point` also works.

Can you elaborate a bit on how/where/why you want to use these wrapper types? Is it still mainly to make it easier to work with Base types that we consider to be GeoInterface compatible objects, and arrays thereof, as mentioned in the first post?

rafaqz:
Reasoning:

1. Tests. We are defining objects everywhere for tests but they are  rarely flexible enough to do real testing of all the edge cases.
2. Types that are missing in packages. Not every package implements everything, and wrappers can be used to fill those gaps
3. Users cobbling together arbitrary geometries into objects from base points or any other geometries - even mixed together. This is currently really hard. How do we just make a `LineString` out of any kind of point? you have to learn it again and again for every package. 
4. Packages cobbling together objects for internal users e.g. MultiPolygons from polygons. How can they currently do that generically for any GeoInterface.jl compatible polygons without a dependency?
5. Creating arbitrary features and feature collections. Not possible in half the packages and very useful to have a general method, see e.g. https://github.com/JuliaGeo/GeoInterface.jl/issues/79

src/fallbacks.jl
rafaqz: The other `AbstractPolygonTrait` types besides `Polygon` don't have rings. Its a pretty bad a bug currently that we treat them like they do.

visr:
Thanks! Would you mind adding this to the docs as well? Perhaps also including when it's not needed/helpful to use these? For packages like GeoJSON and GeometryBasics I guess there is not directly much benefit from these, or not?

rafaqz:
Yeah, I will add that.

Its actually pretty beneficial to GeometryBasics.jl as it has no Feature/FeatureCollection concept, again see this for a use case https://github.com/JuliaGeo/GeoInterface.jl/issues/79.

Also most packages (e.g. Rasters.jl) don't know about these packages or depend on them, so would still use the wrapper type over the package type as its generic. Even if GeoJSON has its own types how would we know about them without a dependency, and why would we write anything package specific when we don't have to.

I imagine it will be the same for users. Just learn one set of wrappers and use them everywhere, even if there are native types why bother learning that.

src/fallbacks.jl
evetion: I think they should behave as single rings? So a Triangle is a Polygon with a single ring, we should not skip directly to points.

src/fallbacks.jl
```diff
@@ -66,6 +71,10 @@ getpoint(t::AbstractMultiPolygonTrait, geom) = flatten((p for p in getpoint(r))
 npatch(t::AbstractPolyhedralSurfaceTrait, geom)::Integer = ngeom(t, geom)
 getpatch(t::AbstractPolyhedralSurfaceTrait, geom) = getgeom(t, geom)
 getpatch(t::AbstractPolyhedralSurfaceTrait, geom, i::Integer) = getgeom(t, geom, i)
+getpoint(t::AbstractPolyhedralSurfaceTrait, geom) = flatten((p for p in getpoint(ls)) for ls in getgeom(t, geom))
```
evetion, line 74: A surface consists of patches (Polygons), not of linestrings (ls).

rafaqz: Shouldn't this just work anyway because the flattening is recursive?

src/fallbacks.jl
rafaqz: Its a bit confusing as in the diagrams on page 14 and 30 of the spec triangles are not built from a LinearRing: 

file:///home/raf/Downloads/06-103r4_Implementation_Specification_for_Geographic_Information_-_Simple_feature_access_-_Part_1_Common_Architecture_v1.2.1.pdf

Do you have a reference for where its specified that they have a single ring?

rafaqz: Bump @evetion this is blocking. I don't think there are any implementations of `Trangle` anywhere yet either.

Is the idea that `getgeom` on a `TriangleTrait` and others will always return a length one iterator of `LinearRingTrait`?

I guess that can also make sense? `getpoint` will still return an iterator over the points so it doesn't make so much difference. One question is if wrappers will require the nesting in constructors - will a vector of `PointTrait` be acceptable input for a `Triangle` or does it have to be vector of vector of `PointTrait`?

rafaqz:
This is good to go, with a section in the docs.

I simplified it to just include the main types needed for e.g. LibGEOS tests. 

We can work through issues (like with how to construct `Triangle`) with the other ones later, best to get this merged before I run out of steam to finish this and the PRs depending on it.

rafaqz:
Coverage is up at 98%, and it seems like everthing makes sense - using it in LibGEOS tests seems to work well.

@visr @evetion can I get an approval to merge this?

evetion:
🎉 I'll try to review tomorrow

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+is3d(::WrapperGeometry{Z}) where Z = Z
+ismeasured(::WrapperGeometry{<:Any,M})  where M = M
+
+Base.parent(geom::WrapperGeometry) = geom.geom
+
+function Base.:(==)(g1::WrapperGeometry, g2::WrapperGeometry)
+    all(((a, b),) -> a == b, zip(getgeom(g1), getgeom(g2)))
+end
+
+geointerface_geomtype(trait) = throw(ArgumentError("trait $trait not yet implemented as a GeoInterface geometry wrapper"))
```
evetion, line 73: It's not covered by tests, but it also doesn't seem to do anything (anymore)? Can be removed?

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+    isgeometry(T) ? getgeom(trait, parent(geom)) : parent(geom)
+end
+
+extent(::AbstractGeometryTrait, geom::WrapperGeometry) =
+    isnothing(geom.extent) && isgeometry(parent(geom)) ? extent(parent(geom)) : geom.extent
+function ngeom(trait::AbstractGeometryTrait, geom::WrapperGeometry{<:Any,<:Any,T}) where T
+    isgeometry(T) ? ngeom(parent(geom)) : length(parent(geom))
+end
+
+for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
```
evetion, line 90: Might be good to explain length_check and nesting for future additions to this list.

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+    isfeaturecollection(parent(fc)) ? nfeature(t, parent(fc)) : length(parent(fc))
+end
+getfeature(::FeatureCollectionTrait, fc::FeatureCollection) =
+    isfeaturecollection(parent(fc)) ? getfeature(t, parent(fc)) : parent(fc)
+getfeature(t::FeatureCollectionTrait, fc::FeatureCollection, i::Integer) =
+    isfeaturecollection(parent(fc)) ? getfeature(t, parent(fc), i) : parent(fc)[i]
+extent(fc::FeatureCollection) =
+    (isnothing(fc.extent) && isfeaturecollection(parent(fc))) ? extent(parent(fc)) : fc.extent
+crs(fc::FeatureCollection) =
+    isfeaturecollection(parent(fc)) && isnothing(fc.crs) ? crs(parent(fc)) : fc.crs
```
evetion, line 440: These all repeat the `isfeaturecollection(parent(fc))` check, which is also done in the constructor. Might be good to generalize.

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+"""
+struct FeatureCollection{P,C,E}
+    parent::P
+    crs::C
+    extent::E
+end
+function FeatureCollection(fc::FeatureCollection; crs=crs(fc), extent=extent(fc))
+    FeatureCollection(parent(fc), crs, extent)
+end
+function FeatureCollection(parent; crs=nothing, extent=nothing)
```
evetion, line 413: One can override the crs and extent of a `parent` here, but those overrides are never returned anymore if `isfeaturecollection(parent)`.

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+```julia
+feature = Feature(geom; properties=(a=1, b="2"), crs=EPSG(4326))
+```
+"""
+struct Feature{T,C,E<:Union{Extents.Extent,Nothing}}
+    parent::T
+    crs::C
+    extent::E
+end
+function Feature(f::Feature; crs=f.crs, extent=f.extent, properties=nothing)
```
evetion, line 352: Here you use `f.crs`, while FeatureCollection below uses `crs(f)`.

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+    p = (; X, Y)
+    if !isnothing(Z)
+        p = merge(p, (; Z))
+    end
+    if !isnothing(M)
+        p = merge(p, (; M))
+    end
+    return Point(p)
+end
+function Point(geom)
```
evetion, line 277: Maybe catch the case where we can pass a Point to a Point and prevent recursive wrappers?

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+
+x(trait::PointTrait, geom::Point) = x(trait, parent(geom))
+y(trait::PointTrait, geom::Point) = y(trait, parent(geom))
+z(trait::PointTrait, geom::Point{true}) = z(trait, parent(geom))
+z(trait::PointTrait, geom::Point{false}) = _no_z_error()
+m(trait::PointTrait, geom::Point{<:Any,false}) = _no_m_error()
+m(trait::PointTrait, geom::Point{<:Any,true}) = m(trait, parent(geom))
+# Special-case vector and tuple points so we can force them to be measured while not 3d
+m(trait::PointTrait, geom::Point{false,true,<:PointTuple3}) = parent(geom)[3]
+m(trait::PointTrait, geom::Point{false,true,<:Vector{<:Real}}) = parent(geom)[3]
```
evetion, line 307: ❤️

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+        $geomtype
+
+        $geomtype(geom, [extent])
+        $geomtype{Z,M}(geom, [extent])
+
+    ## Arguments
+
+    - `geom`: any object returning $trait from `GeoInterface.trait`, or a vector
+        of objects returning $child_trait.
+    - `extent`: an `Extents.Extent`, or `nothing`.
```
evetion, line 126: What about crs?

src/wrappers.jl
```diff
@@ -0,0 +1,442 @@
...
+                        _wrong_child_error($geomtype, $child_trait, child)
+                    end
+                else
+                    # Otherwise compain the nested child type is wrong
+                    _wrong_child_error($geomtype, $child_trait, child)
+                end
+            end
+        else
+            # Or complain the parent type is wrong
+            _parent_type_error(geom)
```
evetion, line 216: Would be good to have a test for this clause.

evetion (commented):
Gargantuan effort, very meta!

evetion:
Closes #75

src/wrappers.jl
rafaqz: I was wondering that too. Should geometries have crs attached? Is there a use case? I have that just for `Feature` and `FeatureCollection` now.

For extents I think there is a use case - shapefile geometries all have a precalculated extent but others formats don't and have no way to add them. These wrappers would provide a way to do that without allocating new objects.

src/wrappers.jl
rafaqz: Good point, yes overrides should work in all cases

src/wrappers.jl
rafaqz: Ok I'll may a method

test/wrappers.jl
```diff
@@ -0,0 +1,251 @@
...
+@test_throws ArgumentError GI.Feature(:not_a_feature; properties=(x=1, y=2, z=3))
+
+# Feature Collection
+fc = GI.FeatureCollection([feature]; crs=EPSG(4326), extent=GI.extent(feature))
+@test fc === GI.FeatureCollection(fc)
+@test GI.crs(fc) == EPSG(4326)
+# TODO this should return 
+@test GI.extent(fc) == fc.extent
+@test first(GI.getfeature(fc)) == GI.getfeature(fc, 1) === feature
+@test GI.testgeometry(multipolygon)
```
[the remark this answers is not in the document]
rafaqz: Not sure what you mean here

rafaqz:
@evetion thanks for the super-thorough review, gives me a lot of confidence in quality of the PR and the interface generally.

It made me think we should aim for 100% coverage and this degree of correctness now that we are not rushing to just get the basics working.

src/wrappers.jl
evetion: It's in the interface to request a crs for a geometry, so it would make sense to include here.

test/wrappers.jl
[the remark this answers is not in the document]
evetion: It tests multipolygon in the section for featurecollection.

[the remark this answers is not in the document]
rafaqz: Oh right, yeah just a typo. I think I find it hard to see mistakes like this where you have correct code but in the wrong place.

src/wrappers.jl
rafaqz: Ok, will do. Do you think even for points?

rafaqz: If not we can merge this

evetion: Sorry, it's hard to distinguish between commit notifications and comments here. Feel free to ping me for bumps. 

I'd say also add crs for Points. If for any geometry, also for Point.

rafaqz:
Ok this has 100% coverage. 

@evetion want to give this one last review? then we can squash and merge. 

(it really needs squashish, there are a lot of trash commits here)

evetion (approved):

PR #84, opened 2022-12-21
fix dispatch on base types and test them

rafaqz:
This fixes `Tuple` and `NamedTuple` points for mixed integers and where `trait` was only looking at `NamedTuple` as a `FeatureTrait` rather than also as a `PointTrait`

See https://github.com/JuliaGeo/Leaflet.jl/issues/6

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1,29 +1,36 @@
 # Implementation of GeoInterface for Base Types
 
+
+const PointTuple2 = Tuple{<:Real,<:Real}
+const PointTuple3 = Tuple{<:Real,<:Real,<:Real}
+const PointTuple4 = Tuple{<:Real,<:Real,<:Real,<:Real}
+const PointTuple = Union{PointTuple2,PointTuple3,PointTuple4}
+
 GeoInterface.isgeometry(::Type{<:AbstractVector{<:Real}}) = true
 GeoInterface.geomtrait(::AbstractVector{<:Real}) = PointTrait()
 GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom)
 GeoInterface.getcoord(::PointTrait, geom::AbstractVector{<:Real}, i) = getindex(geom, i)
 
-GeoInterface.isgeometry(::Type{<:NTuple{N,<:Real}}) where {N} = true
-GeoInterface.geomtrait(::NTuple{N,<:Real}) where {N} = PointTrait()
-GeoInterface.ncoord(::PointTrait, geom::NTuple{N,<:Real}) where {N} = N
-GeoInterface.getcoord(::PointTrait, geom::NTuple{N,<:Real}, i) where {N} = getindex(geom, i)
+GeoInterface.isgeometry(::Type{<:PointTuple}) = true
+GeoInterface.geomtrait(::PointTuple) = PointTrait()
+GeoInterface.ncoord(::PointTrait, geom::PointTuple) = Base.length(geom)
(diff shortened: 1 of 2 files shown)


rafaqz:
@evetion this fixes quite a ew bugs with base points and is needed for fixing another more serious bug in GeoJSON.jl: https://github.com/JuliaGeo/GeoJSON.jl/pull/57 so would be good to get it merged if you have a minute to review

src/base.jl
```diff
@@ -1,30 +1,83 @@
...
 end
 
-# Custom coordinate order/names NamedTuple
-GeoInterface.isgeometry(::Type{<:NamedTuple{Keys,NTuple{N,T}}}) where {Keys,N,T<:Real} = all(in(default_coord_names), Keys)
-GeoInterface.geomtrait(::NamedTuple{Keys,NTuple{N,T}}) where {Keys,N,T<:Real} = PointTrait()
-GeoInterface.ncoord(::PointTrait, geom::NamedTuple{Keys,NTuple{N,T}}) where {Keys,N,T<:Real} = Base.length(geom)
-GeoInterface.getcoord(::PointTrait, geom::NamedTuple{Keys,NTuple{N,T}}, i) where {Keys,N,T<:Real} = getindex(geom, i)
-GeoInterface.coordnames(::PointTrait, geom::NamedTuple{Keys,NTuple{N,T}}) where {Keys,N,T<:Real} = Keys
 
+const NamedTuplePoints = Union{
```
visr, line 29: Do you know if there are any downsides to using such large unions? Is the compiler happy with them? In most cases / WKT this would be the order, though it's nice to support swapped X and Y in at least some cases.

```
XY
XYZ
XYM
XYZM
```

rafaqz: Being a NamedTuple people want to write it however they want and it's weird when e.g.
`(Y=1, X=2)` doesn't work for people - see the linked issue. Then, if one like that works it seems arbitrary and confusing if any other combination doesn't work.

So I figured just have them all. I imagine there is a tiny bit more compilation time with the union, but it's probably negligible? Maybe see if package load time is any longer? I doubt it will be measurable.

src/base.jl
```diff
@@ -1,30 +1,83 @@
...
+}
+
+_keys(::Type{<:NamedTuple{K}}) where K = K
+# Custom coordinate order/names NamedTuple
+GeoInterface.isgeometry(::Type{T}) where {T<:NamedTuplePoints} = all(in(default_coord_names), _keys(T))
+GeoInterface.geomtrait(::NamedTuplePoints) where {Keys} = PointTrait()
+GeoInterface.ncoord(::PointTrait, geom::NamedTuplePoints) where {Keys} = Base.length(geom)
+GeoInterface.getcoord(::PointTrait, geom::NamedTuplePoints, i) where {Keys} = getindex(geom, i)
+GeoInterface.coordnames(::PointTrait, geom::NamedTuplePoints) = _keys(typeof(geom))
+GeoInterface.x(::PointTrait, geom::NamedTuplePoints, i) where {Keys} = geom.X
```
evetion, line 77: What's `{Keys}` here and what it's function? Also, there's no test coverage here, so they might not be needed at all, or we need to expand our coverage.

rafaqz: I've just forgotten to delete it from before

rafaqz: Also good point they may not be needed... maybe the fallbacks handle this fine?

rafaqz:
This is good to go

visr (approved):
Looks pretty clear, with good tests.

PR #89, opened 2023-01-17
fix performance and correctness of base points, and test

rafaqz:
Some performance and correctness fixes. As much as possible we check that tuples and vectors are the right length.

I think the performance of GeoJSON was also affected by how slow the Vector methods were here - they are 100x faster now.

Found working on #78. Lets get this merged quickly so that can be merged

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1,47 +1,89 @@
 # Implementation of GeoInterface for Base Types
 
 
+# AbstractVector{<:Real} length 2 - 4 (where length checking is possible and efficient)
+
+GeoInterface.isgeometry(::Type{<:AbstractVector{<:Real}}) = true
+GeoInterface.geomtrait(::AbstractVector{<:Real}) = PointTrait()
+GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) # TODO should this error for length > 4 ?
+GeoInterface.getcoord(::PointTrait, geom::AbstractVector{<:Real}, i) = getindex(geom, i)
+GeoInterface.is3d(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) in (3, 4)
+GeoInterface.ismeasured(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) == 4 # Measured must have Z
+
+# x/y/z/m methods were 100x slower without these custom definitions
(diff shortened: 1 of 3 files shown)


rafaqz:
Note: needs some tests for `is3d`, `ismeasured` etc on Tuple points

src/base.jl
```diff
@@ -1,47 +1,89 @@
...
-const PointTuple = Union{PointTuple2,PointTuple3,PointTuple4}
+# AbstractVector{<:Real} length 2 - 4 (where length checking is possible and efficient)
 
 GeoInterface.isgeometry(::Type{<:AbstractVector{<:Real}}) = true
 GeoInterface.geomtrait(::AbstractVector{<:Real}) = PointTrait()
-GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom)
+GeoInterface.ncoord(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) # TODO should this error for length > 4 ?
 GeoInterface.getcoord(::PointTrait, geom::AbstractVector{<:Real}, i) = getindex(geom, i)
+GeoInterface.is3d(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) in (3, 4)
+GeoInterface.ismeasured(::PointTrait, geom::AbstractVector{<:Real}) = Base.length(geom) == 4 # Measured must have Z
```
evetion, line 11: It's good to document this somewhere, as XYM is allowed. But I've never seen M geometries in the wild, so if this makes the code way faster, let's do so.

rafaqz: Its more for correctness (but totally I never see them either).

There is no way to use `Gi.z` and `GI.m` function on a length 3 `Tuple` or `Vector` unless we say it is always x/y/z. 

I'm not sure what to do about this really. I will make a workaround in the wrappers PR where the wrapper will determine which value is Z and M.

evetion: I think what you've done is fine here, XYZ makes more sense than XYM if we have to choose one for 3 length non-named items. But I do like to see this assumption documented somewhere.

rafaqz: Sure, I'll write some docs.

rafaqz: I've added docs to the x/y/z/m methods clarifying point behaviour.

rafaqz:
This should be ready to merge

evetion (approved):

PR #90, opened 2023-01-17
fallback for crs with no geotrait

joshday:
For `x` such that `geomtrait(x) === nothing`, `crs(x)` results in an error.  This PR provides a fallback that returns `nothing` instead.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -94,7 +94,7 @@ issimple(t::AbstractMultiPointTrait, geom) = allunique((getgeom(t, geom)))
 issimple(t::AbstractMultiCurveTrait, geom) = all(issimple.(getgeom(t, geom)))
 isclosed(t::AbstractMultiCurveTrait, geom) = all(isclosed.(getgeom(t, geom)))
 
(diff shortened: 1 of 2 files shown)


rafaqz:
Check `GeoInterface isgeometry` before anything is the alternative to this I guess.

I'm just wondering if this will trigger a bunch of other methods needing to handle `nothing`.

joshday:
> I'm just wondering if this will trigger a bunch of other methods needing to handle nothing.

I did not think this through...I really don't want to kick off a bunch of new fallback methods.  I'm gonna close this.  People can check `geomtrait(x) !== nothing`.

rafaqz:
`isgeometry` should be the preffered check as it always compiles away.

joshday:
Well wouldn't you also need to check `isfeature` and `isfeaturecollection`?

rafaqz:
not if you want the answer to `geomtrait(x) !== nothing`

joshday:
Ah, I was thinking `geomtrait` was used to get e.g. `FeatureCollectionTrait`, but it's just `trait`

PR #93, opened 2023-02-06
add LinearRing recipes and test plotting, also test LineString

rafaqz:
Found linear ring had no recipe while working on https://github.com/JuliaGeo/Shapefile.jl/pull/85

I'm not sure how this worked before for `LinearRing`, it already had tests but were failing on 1.9 for me.

I also tested `LineString` while fixing this.


visr (approved):
Nice, thanks!

PR #94, opened 2023-02-07
fall back to Extents.extent

rafaqz:
@evetion what do you think of falling back to `Extents.extent` ?

It would solve e.g. this issue https://github.com/rafaqz/Extents.jl/issues/7

Also means packages can define `Extents.extent` instead of `GeoInterface.extent`. So the difference would be that `GeoInterface.extent` calculates the extent if it doesn't exist because it knows how to do that with geometries.

That may have solved my earlier issue with always doing the calculations, wish I thought about this more then...

### DIFF
--- a/docs/src/guides/developer.md
+++ b/docs/src/guides/developer.md
@@ -37,6 +37,10 @@ GeoInterface.crs(geomtrait(geom), geom::customgeom)::GeoFormatTypes.GeoFormat
 GeoInterface.extent(geomtrait(geom), geom::customgeom)::Extents.Extent
 ```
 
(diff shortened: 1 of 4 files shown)


evetion:
This sorta makes sense, but does it imply we also need something similar for crs/GeoFormatTypes?

This PR requires a documentation update.

rafaqz:
GeoFormatTypes.jl doesn't export any method like `extent`, although it could export `crs`. 

But we don't have to do that now?

evetion:
It's not for this PR, was just triggered by it as it's the only other thing in which we point to another package to fix something.

visr (approved):

PR #95, opened 2023-03-12
Use map instead of collect with a broadcast.

evetion:
Over at https://github.com/JuliaGeo/LibGEOS.jl/issues/160, this makes `@code_warntype` green for `coordinates(polygon)` and changes the type for coordinates on a MultiPolygon from Any to a Vector.


rafaqz (approved):
Probably faster compilation too.

PR #96, opened 2023-04-29
Tweak extent fallback

rafaqz:
This PR lets non-geometries pass through to Extents.jl if used in `GeoInterface.extent`. Just a convenience really so the function here just works in more cases.

I also used `import` rather than `using` at the same time for correctness.

### DIFF
--- a/src/GeoInterface.jl
+++ b/src/GeoInterface.jl
@@ -1,7 +1,6 @@
 module GeoInterface
 
-import Extents
-using Extents: Extent
+using Extents: Extents, Extent
 using Base.Iterators: flatten
 
 export testgeometry, isgeometry, trait, geomtrait, ncoord, getcoord, ngeom, getgeom
--- a/src/GeoInterface.jl
+++ b/src/GeoInterface.jl
@@ -1,6 +1,7 @@
 module GeoInterface
 
-using Extents
+import Extents
+using Extents: Extent
 using Base.Iterators: flatten
 
 export testgeometry, isgeometry, trait, geomtrait, ncoord, getcoord, ngeom, getgeom
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -107,9 +107,9 @@ function coordinates(t::AbstractFeatureTrait, feature)
     geom = geometry(feature)
     isnothing(geom) ? [] : coordinates(geom)
 end
-coordinates(t::AbstractFeatureCollectionTrait, fc) = [coordinates(f) for f in getfeature(t, fc)]
+coordinates(t::AbstractFeatureCollectionTrait, fc) = map(f -> coordinates(f), getfeature(t, fc))
 
-extent(::AbstractTrait, x) = Extents.extent(x)
+extent(::Any, x) = Extents.extent(x)
 function calc_extent(t::AbstractPointTrait, geom)
     coords = collect(getcoord(t, geom))
     names = coordnames(geom)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -423,9 +423,10 @@ When `fallback` is true, and the obj does not have an extent,
 an extent is calculated from the coordinates of all geometries in `obj`.
 """
 function extent(obj; fallback=true)
-    isnothing(trait(obj)) && return Extents.extent(obj)
-    ex = extent(trait(obj), obj)
-    isnothing(ex) && fallback && return calc_extent(trait(obj), obj)
+    t = trait(obj)
+    isnothing(t) && return Extents.extent(obj)
+    ex = extent(t, obj)
+    isnothing(ex) && fallback && return calc_extent(t, obj)
     return ex
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -423,6 +423,7 @@ When `fallback` is true, and the obj does not have an extent,
 an extent is calculated from the coordinates of all geometries in `obj`.
 """
 function extent(obj; fallback=true)
+    isnothing(trait(obj)) && return Extents.extent(obj)
     ex = extent(trait(obj), obj)
     isnothing(ex) && fallback && return calc_extent(trait(obj), obj)
     return ex
--- a/test/test_primitives.jl
+++ b/test/test_primitives.jl
@@ -489,6 +489,6 @@ Extents.extent(::ExtentPolygon) = Extent(X=(1, 2), Y=(3, 4))
 
 @testset "Extents.jl extent fallback" begin
     @test GeoInterface.extent(ExtentPolygon()) == Extent(X=(1, 2), Y=(3, 4))
-    @test GeoInterface.extent(1) == nothing
-    @test GeoInterface.extent(nothing, 1) == nothing
+    @test isnothing(GeoInterface.extent(1))
+    @test isnothing(GeoInterface.extent(nothing, 1))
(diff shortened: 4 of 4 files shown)


src/GeoInterface.jl
```diff
@@ -1,6 +1,7 @@
 module GeoInterface
 
-using Extents
+import Extents
+using Extents: Extent
```
visr, line 4: I usually combine the two like this in one line:

```suggestion
using Extents: Extents, Extent
```

src/interface.jl
```diff
@@ -423,6 +423,7 @@ When `fallback` is true, and the obj does not have an extent,
 an extent is calculated from the coordinates of all geometries in `obj`.
 """
 function extent(obj; fallback=true)
+    isnothing(trait(obj)) && return Extents.extent(obj)
     ex = extent(trait(obj), obj)
     isnothing(ex) && fallback && return calc_extent(trait(obj), obj)
```
visr, line 428: We call `trait(obj)` 3 times on 3 consecutive lines. Perhaps better to assign it to a variable once.

test/test_primitives.jl
```diff
@@ -489,4 +489,6 @@ Extents.extent(::ExtentPolygon) = Extent(X=(1, 2), Y=(3, 4))
 
 @testset "Extents.jl extent fallback" begin
     @test GeoInterface.extent(ExtentPolygon()) == Extent(X=(1, 2), Y=(3, 4))
+    @test GeoInterface.extent(1) == nothing
+    @test GeoInterface.extent(nothing, 1) == nothing
```
visr, line 493:
```suggestion
    @test isnothing(GeoInterface.extent(1))
    @test isnothing(GeoInterface.extent(nothing, 1))
```

visr (commented):
Seems fine to me. A few minor comments.

src/interface.jl
rafaqz: It compules away ;)

But yeah thas cleaner

src/fallbacks.jl
```diff
@@ -107,9 +107,9 @@ function coordinates(t::AbstractFeatureTrait, feature)
     geom = geometry(feature)
     isnothing(geom) ? [] : coordinates(geom)
 end
-coordinates(t::AbstractFeatureCollectionTrait, fc) = [coordinates(f) for f in getfeature(t, fc)]
+coordinates(t::AbstractFeatureCollectionTrait, fc) = map(f -> coordinates(f), getfeature(t, fc))
```
evetion, line 110:
```suggestion
coordinates(t::AbstractFeatureCollectionTrait, fc) = map(coordinates, getfeature(t, fc))
```

src/fallbacks.jl
```diff
@@ -107,9 +107,9 @@ function coordinates(t::AbstractFeatureTrait, feature)
...
 
-extent(::AbstractTrait, x) = Extents.extent(x)
```
evetion, line 112: We catch the possible `nothing` trait in the other method, so this change isn't needed?

src/interface.jl
```diff
@@ -423,8 +423,10 @@ When `fallback` is true, and the obj does not have an extent,
 an extent is calculated from the coordinates of all geometries in `obj`.
 """
 function extent(obj; fallback=true)
-    ex = extent(trait(obj), obj)
-    isnothing(ex) && fallback && return calc_extent(trait(obj), obj)
+    t = trait(obj)
+    isnothing(t) && return Extents.extent(obj)
```
evetion, line 427: Can we document the possibilities here? Before we probably errored, when the trait was `nothing`, but now the object is passed to Extents, which has its own fallback(?) to nothing?
Does it make sense to just return nothing instead ourselves, as we're asking `GeoInterface.extent(obj)` for an extent, not `Extents.extent(obj)`?

src/fallbacks.jl
rafaqz: The idea here is that `Extents.extent` is more generic and will get the extent of a other objects like a Raster or YAXarray if they happen to be passed to `GeoInterface.extent`

evetion: Then it relates it to https://github.com/JuliaGeo/GeoInterface.jl/pull/76? Might be good to revive it, without the whole affine discussion, just have a trait/extent for now?

rafaqz: No its unrelated to #76 

This is just so that you can use any object as a bounding box anywhere that that information is available.

In Rasters.jl I have to handle this distinction in a bunch of unnessesary places, which means everyone else will have to as well.

src/interface.jl
rafaqz: They both return and `Extent` or `nothing`, and packages can depend on Extents.jl and not GeoInterface.jl

I can't see why we wouldn't fall back to the Extents.jl method

evetion: I have never implemented Extents.Extent directly, only via GeoInterface.extent, so this behavior was surprising to me.

I think this is mostly a question of design and clarity, if you use GeoInterface, you either get something if a package implements GeoInterface, or nothing. In select cases, we can provide a fallback by calculating something. Here you propose to also fallback to other packages (which can return even other things, or error?).

Essentially, how to combine the GeoInterface world with the non-GeoInterface world? As you know, I have a slight preference for the world where more packages implement GeoInterface, and we extend from having just `isgeometry` to GeoInterface.issomethingthatsupportspartsofGeoInterface(Type{x}) == true. Like `israster,` or even just `is`/`hasspatial`, so we can nicely dispatch on extent, crs and the like.

src/fallbacks.jl
evetion: Can you link to that code so I can understand the issue a bit better? I don't want to hold up a pragmatic fix over pedantic architecture discussions, although I do like a clean architecture.

On a related note, I'm implementing GeoInterface for Extent.

src/interface.jl
rafaqz: Currently YAXArrays and Rasters get `Extents.extent` for free via DimensionalData.jl dependency. YAXArrays does not have a GeoInterface dependency.

`Tyler.Map` also defines `Extents.extent` directly:
https://github.com/MakieOrg/Tyler.jl/blob/9d32abe9f25a77d774f6b5d49f2f24667ef0e496/src/Tyler.jl#L179

My design goal is to make it easy to pass as wide as possible variety of objects to any GeoInterace related function that needs an extent/bounding box, facilitated by `GeoInterface.extent` either retrieving or calculating extents for geometries and other GeoInterface impelemting objects, or falling back to `Extents.extent` for other packages that do not implement GeoInterface.

This is easier than saying "Just implement GeoInterface". Because Extents.jl is not just some random "other package", its the package that defines the `Extent` object that we already depend on, and the change is tiny to make here as is clear in this PR.

Saying that packages either implement `GeoInterface.extent` or you get `nothing` rather than the known `Extent` they return from `Extents.extent` seems a little pedantic to me. It also means packages have to implement both `Extents.extent` and `GeoInterface.extent`. 

For some code, my specific need for this and the other extents PR is to reduce this extent handling mess:
https://github.com/rafaqz/Rasters.jl/blob/main/src/utils.jl#L252-L318

Where you can see both `Extents.extent` and `GeoInterface.extent` used on different objects.

evetion (approved):

PR #97, opened 2023-04-29
allow LineStringTrait in Polygon wrapper

rafaqz:
Some packages don't have `LinearRing` types and builld `Polygon` from `LineString`, so that should be allowed here.


rafaqz:
I'm merging this, it's is breaking GeometryBasics.jl interop.

PR #98, opened 2023-04-29
Type stable calc extent

rafaqz:
@asinghvi17 hit some very slow code with Rasters.jl and GeometryBasics.jl where I had forgotten to use `fallbacks=false`.

But it seems it can be a lot faster by using `is3d` to in worst case get small union optimisations (but mostly compiled away), getting the extrema over all points in a geometry in one pass, and avoiding collecting points or making `NamedTuple` at the leaf nodes of the geometry.


evetion:
This looks nice! Can we add a test for the type stability (or hit the missing coverage)?

rafaqz:
Its a small union optimisation, so its not guaranteed to be inferred, just fast.

I'll try to get the code coverage up

rafaqz:
Pretty weird that this can work on 1.9 and not 1.6... will have to see where there could be something different

rafaqz:
@evetion this is good to go if you could approve, coverage is passing

src/fallbacks.jl
```diff
@@ -111,11 +111,27 @@ coordinates(t::AbstractFeatureCollectionTrait, fc) = [coordinates(f) for f in ge
...
+    X = extrema(p -> x(p), points)
+    Y = extrema(p -> y(p), points)
+    if is3d(geom)
+        Z = extrema(p -> z(p), points)
+        Extent(; X, Y, Z)
+    else
+        Extent(; X, Y)
+    end
+end
+calc_extent(t::GeometryCollectionTrait, geom) = reduce(Extents.union, (extent(f) for f in getgeom(t, geom)))
```
evetion, line 134: pedantry: note that we call `extent` (and thus accept pre-calculated extents) for multigeoms instead of calling `calc_extent` (recalculating everything). Old behaviour and quite pragmatic, but it might be slightly unexpected.

evetion (approved):

PR #105, opened 2023-07-23
Add WellKnownGeometry as a package extension.

evetion:
So there's a fallback for `asbinary` and `astext` once WellKnownGeometry is loaded.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,14 +6,22 @@ version = "1.3.1"
 [deps]
 Extents = "411431e0-e8b7-467b-b5e0-f676ba4f2910"
 
+[weakdeps]
+WellKnownGeometry = "0f680547-7be7-4555-8820-bb198eeb646b"
+
+[extensions]
+WellKnownGeometryExt = "WellKnownGeometry"
+
 [compat]
 Extents = "0.1.1"
+WellKnownGeometry = "0.2"
 julia = "1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 GeoFormatTypes = "68eda718-8dee-11e9-39e7-89f7f65f511f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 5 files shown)


felixcremer:
Don't we get a circular dependency problem when GeoInterface depends on WellKnownGeometry and vice versa?

evetion:
> Don't we get a circular dependency problem when GeoInterface depends on WellKnownGeometry and vice versa?

If we add it directly yes. Indirect, I'm not sure (could explain the errors/warnings), but it does work. Would be neat to use this pattern for future package extensions, for example LibGEOS, providing a fallback things like `intersection(geomA, geomB)`.

rafaqz:
The circular dependency sounds like a problem, what happens when you need to bump the version?

With the LibGEOS idea, you can already do `LibGEOS.intersection(anygeom, anygeom)` and it works on any GeoInterface compatible geometries.

I've been thinking it makes more sense to just align all the function names in LibGEOS and other packages to the names used here, rather than having a fallback like that here... Then users are explicitly choosing the backend.

evetion:
Well apart from circularity (we control both compats, so shouldn't be a big issue), this is about a design to handle (multiple) backends.

ArchGDAL, LibGEOS and WellKnownGeometry all could be the backend here, but implementing GeoInterface.asbinary(:Any) in a package is piracy and loading multiple packages at the same time will clash.

Instead of saying X is the blessed package by only implementing it there, I rather be explicit in GeoInterface, having all blessed fallbacks defined here. Package extensions seemed to fit the bill, but maybe not.

felixcremer:
Could we move this extension to live in WellKnownGeometry?  That  way we would get the same behaviour of loading this code when both GeoInterface and WellKnownText are specifically loaded, but without the circular dependency because WellKnownText depends on GeoInterface anyways. And we could do the same for the other backends, the only problem is, that we would get clashes, when we load two backends, but this is something that would also happen when we put the extensions here.

rafaqz:
Yeah isnt the circular thing is a problem for CI? Like how can tests pass for one without the other when you bump compat?

PR #108, opened 2023-08-04
Use GI.convert for Makie plots, allow arrays

rafaqz:
@asinghvi17 I took the liberty of making a PR from your branch, I have this array code somewhere too but we may as well merge this instead. `Float64` is also better for high resolution things, going on the FLoat32 artifacts we see at high zooms in Tyler.jl....

~~A test of plotting the vectors could help I guess~~ done

### DIFF
--- a/GeoInterfaceMakie/src/GeoInterfaceMakie.jl
+++ b/GeoInterfaceMakie/src/GeoInterfaceMakie.jl
@@ -18,13 +18,6 @@ end
 function plottype_from_geomtrait(::Union{GI.PolygonTrait,GI.MultiPolygonTrait, GI.LinearRingTrait})
     MC.Poly
 end
-function pttype(geom)
-    if GI.is3d(geom)
-        GB.Point3{Float64}
-    else
-        GB.Point2{Float64}
-    end
-end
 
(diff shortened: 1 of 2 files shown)


rafaqz:
@jw3126 I swapped this over to just using `GeoInterface.convert(GeometryBasics, geom)` and removing all the logic here, let me know if this makes sense or has broken anything.

Arrays of geometries are working fine now (except where they don't work with GeometryBasics either) but the plot type detection isn't working - and array of polygons is plotted as a scatter with `plot`. Any ideas about how to fix that @asinghvi17 ?

jw3126:
> @jw3126 I swapped this over to just using `GeoInterface.convert(GeometryBasics, geom)` and removing all the logic here, let me know if this makes sense or has broken anything.

I saw you skipped some tests, were these broken before, or does this PR break them?

rafaqz:
I removed `points` as its not used anymore, so I also removed the tests... but wasn't sure if anyone was using it manually for anything? 

You could instead use `GI.convert.(Ref(GeometryBasics), GI.getpoint(geom))`

jw3126 (approved):
Ok cool LGTM thanks!

PR #116, opened 2023-12-21
Add crs is nothing fallback for non-geointerface types

rafaqz:
This PR loosens the type in the `crs` fallback so that `crs(x) == nothing` for non geointerface `x`.

It also tests this fallback for both `crs` and `extent`

edit:
```julia
# Current behaviour
julia> GeoInterface.crs(1)
ERROR: MethodError: no method matching crs(::Nothing, ::Int64)
```


evetion:
Might be good to distinguish between the two cases?
```julia
crs(::Nothing, nongeom) = nothing
crs(::AbstractTrait, geom) = gft"engineeringcrs/unknowncrs"
```

rafaqz:
Its because in Rasters I just call `crs` on anything and handle the `nothing` return value. Easier than checking if there us a `crs` method defined.

So trying yo swap in `GeoInterface.crs` causes a lot of problems

rafaqz:
Ok its separated into two lines now

evetion (approved):

PR #117, opened 2023-12-27
allow missing in arrays for Makie plots

rafaqz:
Minor change to allow `missing` in GeoInterfaceMakie plots. 

This is needed for Shapefile.jl makie recipes.


jw3126 (approved):
LGTM, should we bump the version?

PR #130, opened 2024-04-21
Add a 1-arg curried version of `GI.convert`

asinghvi17:
I was doing some data processing and having `|> GI.convert(LibGEOS) |> ...` would have been handy, so I coded this up.


rafaqz:
Seems useful. Unfortunate they are both kind of slow when we are doing repeat applications.

rafaqz (approved):

visr:
Looks good to me, though we should add a test for these two new methods.

visr (approved):

PR #136, opened 2024-06-25
Allow vectors with mixed single and multi geometries to be plotted in Makie

asinghvi17:
This upstreams GeoMakie.to_multipoly and its variants into GeoInterfaceMakie.  Performance needs improvement but this is otherwise ready for review, and I have tested it works.

Usecase:

```julia
using NaturalEarth
using CairoMakie

fc = NaturalEarth.naturalearth("admin_0_countries", 10)
plot(fc.geometry)
```
without any fuss or muss.

### DIFF
--- a/GeoInterfaceMakie/src/GeoInterfaceMakie.jl
+++ b/GeoInterfaceMakie/src/GeoInterfaceMakie.jl
@@ -144,6 +144,12 @@ to_multipoly(::Nothing, geom::AbstractVector) = to_multipoly.(GeoInterface.trait
 to_multipoly(::GeoInterface.PolygonTrait, geom) = GB.MultiPolygon([GeoInterface.convert(GB, geom)])
 to_multipoly(::GeoInterface.MultiPolygonTrait, geom) = GeoInterface.convert(GB, geom)
 
+function to_multipoly(::GeoInterface.GeometryCollectionTrait, geom)
+    ls_or_mls = filter(x -> GI.geomtrait(x) isa Union{GI.MultiPolygonTrait, GI.PolygonTrait}, GI.getgeom(geom))
+    multipolys = to_multipoly(ls_or_mls)
+    return GeometryBasics.MultiPolygon(vcat(getproperty.(multipolys, :polys)...))
+end
+
 to_multilinestring(poly::GB.LineString) = GB.MultiLineString([poly])
 to_multilinestring(poly::Vector{GB.Polygon}) = GB.MultiLineString(poly)
 to_multilinestring(mp::GB.MultiLineString) = mp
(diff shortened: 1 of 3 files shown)


GeoInterfaceMakie/src/GeoInterfaceMakie.jl
```diff
@@ -23,12 +23,32 @@ function _convert_arguments(t, geom)::Tuple
...
         geob = map(geom -> GI.convert(GB, geom), skipmissing(geoms))
     else
         geob = map(geom -> GI.convert(GB, geom), geoms)
     end
+    if !(eltype(geob) <: GB.AbstractGeometry) || eltype(geob) isa Union # Unions are bad
+        if isempty(geob)
+            geob = geob
+        end
+        first_trait = GI.geomtrait(first(geob))
+        last_trait = GI.geomtrait(last(geob))
```
rafaqz, line 38: What if the first and last are the same but the middle is different?

asinghvi17: We already know that something is different because the types are not the same.  But a more robust check would be nice.

asinghvi17:
Currently this only allows vectors of (poly, multipoly) or (linestring, multilinestring).  Should we also allow GeometryCollections, and just assume that GeoMakie or GeometryOps will do some piracy to allow plotting those, maybe an error hinter or something?

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
```diff
@@ -23,12 +23,30 @@ function _convert_arguments(t, geom)::Tuple
...
+        end
+        first_trait = GI.geomtrait(first(geob))
+        different_trait_idx = findfirst(x -> GI.geomtrait(x) != first_trait, geob)
+        if isnothing(different_trait_idx)
+            # All traits are the same, and all things are GB, so how did this happen?
+            # It's possible that the point lengths are different, we still need to handle that.
+            return MC.convert_arguments(t, geob)
+        end
+        if all(x -> GI.geomtrait(x) isa Union{GI.MultiPolygonTrait, GI.PolygonTrait, GI.GeometryCollectionTrait}, geob) # an array of polygon like structs
+            geob = to_multipoly(geob)
```
rafaqz, line 45: Might generate better code if you return here

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
```diff
@@ -23,12 +23,30 @@ function _convert_arguments(t, geom)::Tuple
...
+        elseif all(x -> GI.geomtrait(x) isa Union{GI.MultiLineStringTrait, GI.LineStringTrait, GI.GeometryCollectionTrait}, geob) # an array of linestring like structs
+            geob = to_multilinestring(geob)
```
rafaqz, line 47: And here

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
asinghvi17:
```suggestion
            return to_multipoly(geob)

GeoInterfaceMakie/src/GeoInterfaceMakie.jl
asinghvi17:
```suggestion
        if all(x -> GI.geomtrait(x) isa Union{GI.MultiPolygonTrait, GI.PolygonTrait, GI.GeometryCollectionTrait}, geob) # an array of polygon like structs
            return MC.convert_arguments(t, to_multipoly(geob))
        elseif all(x -> GI.geomtrait(x) isa Union{GI.MultiLineStringTrait, GI.LineStringTrait, GI.GeometryCollectionTrait}, geob) # an array of linestring like structs
            return MC.convert_arguments(t, to_multilinestring(geob))

rafaqz:
Not sure GeometryOps should get involved here, that's some messy piracy... Why not deal with GeometryCollection here too?

asinghvi17:
Hmm I guess we can assume gc's are well formed...

00krishna:
Do you need any additional input from me @asinghvi17 ? I am good with whatever you choose.

asinghvi17:
I think this approach is good, only problem is that GeometryBasics doesn't support geometry collections.  I guess I'd have to rewrite the logic to check for traits then dispatch accordingly instead of eagerly converting.

rafaqz:
Can we just convert to vector and hit the other vector methods in this PR?

asinghvi17:
What do you mean?  Just skipping GeometryCollection support?

rafaqz:
No, just collect the geometry collection and call the method again to get Vector dispatch and the Union branch you defined in this PR

asinghvi17:
For the plot type that ought to work - the problem is that there's no GB convert defined, so the convert_arguments for vectors will fail.  Just have to reorder the evaluations and check traits first...

PR #140, opened 2024-07-04
Implement `crstrait`.

### LINKED ISSUE #119: Add CRS traits
This is a design issue, feel free to comment and/or come up with your own thoughts.

I would like to be able to distinguish between spherical and cartesian geometries and coordinate systems. For example, if you currently calculate (using Geomorphology) the slope on a geographic raster the answer is **wrong**. If you currently calculate the area of a geographic geometry, it is **wrong**. Libraries like LibGEOS and others assume cartesian coordinates, but don't (and can't) error or warn for geographic geometries. This is something we extensively discussed at https://r-spatial.org/sdsl/.

GeoInterface seems the correct place to implement something to improve the state of the ecosystem. Other libraries distinguish between Geography and Geometry types (PostGIS), others solve it purely using the attached coordinate reference system. I propose to use traits to distinguish between the two modes, forcing libraries to choose what to implement.

```julia
abstract type AbstractCRSTrait end
struct ProjectedTrait <: AbstractCRSTrait end
struct GeographicTrait <: AbstractCRSTrait end

GeoInterface.crstrait(geom) = crstrait(crs(geom))  # One of ProjectedTrait() or GeographicTrait()
crstrait(any::Any) = ProjectedTrait()  # by default

# all relevant methods, example for `area`
area(geom) = area(crstrait(geom), geomtrait(geom), geom)

# for current implementations
@deprecate area(::ProjectedTrait, geomtrait(geom), geom) area(geomtrait(geom), geom)

area(::GeographicTrait, geomtrait(geom), geom) = MethodError("Please use package X to enable geographic calculations.")
```

This design is backwards compatible, defaulting to ProjectedTrait(), but warns if you are not explicit in your implementations. Once a geometry/crs gains the GeographicTrait, you get an error instead of wrong answer.

evetion:
To distinguish between *geographic* and *projected* coordinate systems. Fixes #119.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -111,7 +111,7 @@ end
 coordinates(t::AbstractFeatureCollectionTrait, fc) = map(f -> coordinates(f), getfeature(t, fc))
 
 extent(::Any, x) = Extents.extent(x)
-function calc_extent(::AbstractProjectedTrait, t::AbstractPointTrait, geom)
+function calc_extent(t::AbstractPointTrait, geom)
     x = GeoInterface.x(t, geom)
     y = GeoInterface.y(t, geom)
     if is3d(geom)
@@ -121,7 +121,7 @@ function calc_extent(::AbstractProjectedTrait, t::AbstractPointTrait, geom)
         return Extent(; X=(x, x), Y=(y, y))
     end
 end
-function calc_extent(::AbstractProjectedTrait, t::AbstractGeometryTrait, geom)
+function calc_extent(t::AbstractGeometryTrait, geom)
     points = getpoint(t, geom)
     X = extrema(p -> x(p), points)
     Y = extrema(p -> y(p), points)
@@ -132,12 +132,12 @@ function calc_extent(::AbstractProjectedTrait, t::AbstractGeometryTrait, geom)
         Extent(; X, Y)
     end
 end
-calc_extent(::AbstractProjectedTrait, t::GeometryCollectionTrait, geom) = reduce(Extents.union, (extent(f) for f in getgeom(t, geom)))
-function calc_extent(::AbstractProjectedTrait, ::AbstractFeatureTrait, feature)
+calc_extent(t::GeometryCollectionTrait, geom) = reduce(Extents.union, (extent(f) for f in getgeom(t, geom)))
+function calc_extent(::AbstractFeatureTrait, feature)
     geom = geometry(feature)
     isnothing(geom) ? nothing : extent(geom)
 end
-calc_extent(::AbstractProjectedTrait, t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
+calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
 
 # Package level `GeoInterface.convert` method
 # Packages must implement their own `traittype` method
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -111,7 +111,7 @@ end
 coordinates(t::AbstractFeatureCollectionTrait, fc) = map(f -> coordinates(f), getfeature(t, fc))
 
 extent(::Any, x) = Extents.extent(x)
-function calc_extent(t::AbstractPointTrait, geom)
+function calc_extent(::AbstractProjectedTrait, t::AbstractPointTrait, geom)
     x = GeoInterface.x(t, geom)
     y = GeoInterface.y(t, geom)
     if is3d(geom)
@@ -121,7 +121,7 @@ function calc_extent(t::AbstractPointTrait, geom)
         return Extent(; X=(x, x), Y=(y, y))
     end
 end
-function calc_extent(t::AbstractGeometryTrait, geom)
+function calc_extent(::AbstractProjectedTrait, t::AbstractGeometryTrait, geom)
     points = getpoint(t, geom)
     X = extrema(p -> x(p), points)
     Y = extrema(p -> y(p), points)
@@ -132,12 +132,12 @@ function calc_extent(t::AbstractGeometryTrait, geom)
         Extent(; X, Y)
     end
 end
-calc_extent(t::GeometryCollectionTrait, geom) = reduce(Extents.union, (extent(f) for f in getgeom(t, geom)))
-function calc_extent(::AbstractFeatureTrait, feature)
+calc_extent(::AbstractProjectedTrait, t::GeometryCollectionTrait, geom) = reduce(Extents.union, (extent(f) for f in getgeom(t, geom)))
+function calc_extent(::AbstractProjectedTrait, ::AbstractFeatureTrait, feature)
     geom = geometry(feature)
     isnothing(geom) ? nothing : extent(geom)
 end
-calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
+calc_extent(::AbstractProjectedTrait, t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
 
 # Package level `GeoInterface.convert` method
 # Packages must implement their own `traittype` method
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -201,8 +201,8 @@ isring(geom) = isclosed(geom) && issimple(geom)
 Return the length of `geom` in its 2d coordinate system.
 Note that this is only valid for [`AbstractCurveTrait`](@ref)s.
 """
-length(geom) = length(crstrait(geom), geomtrait(geom), geom)
-length(::UnknownTrait, trait, geom) = length(trait, geom)  # fallback
+length(geom) = length(geomtrait(geom), geom)
+
 
 # Surface
 
@@ -212,8 +212,8 @@ length(::UnknownTrait, trait, geom) = length(trait, geom)  # fallback
 Return the area of `geom` in its 2d coordinate system.
 Note that this is only valid for [`AbstractSurfaceTrait`](@ref)s.
 """
-area(geom) = area(crstrait(geom), geomtrait(geom), geom)
-area(::UnknownTrait, trait, geom) = area(trait, geom)  # fallback
+area(geom) = area(geomtrait(geom), geom)
+
 
 """
     centroid(geom) -> Point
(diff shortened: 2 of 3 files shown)


asinghvi17:
I think we would also want similar traits in GFT or some kind of trait wrapper, so that users can choose to have CRS traits evaluated at compile time too!  It should be possible to detect this using Proj I think.

evetion:
> I think we would also want similar traits in GFT or some kind of trait wrapper, so that users can choose to have CRS traits evaluated at compile time too! It should be possible to detect this using Proj I think.

Agreed, this is would be the setup for that. For now, the default UnknownTrait() should be compiled away.

src/interface.jl
```diff
@@ -201,16 +201,19 @@ isring(geom) = isclosed(geom) && issimple(geom)
...
+
 """
     area(geom) -> Number
 
 Return the area of `geom` in its 2d coordinate system.
 Note that this is only valid for [`AbstractSurfaceTrait`](@ref)s.
 """
-area(geom) = area(geomtrait(geom), geom)
+area(geom) = area(crstrait(geom), geomtrait(geom), geom)
+area(::UnknownTrait, trait, geom) = area(trait, geom)  # fallback
```
asinghvi17, line 216: Should this be AbstractProjectedTrait too?

evetion: I had that first, but as soon as a package implements area with ProjectedTrait, this fallback is not needed.

asinghvi17: I guess this goes back to "which is the blessed implementation", though - do we want packages dispatching on their own geometry types?

evetion: I don't think blessed has anything to do with this. To implement GeoInterface, packages already have to dispatch on their own geometry type, it's part of the spec?

rafaqz: But most geometry types are not going to implement these `area` algs right? Like `GeometryOps.area` already works for anything. 

I don't really get the use-case for these traits and `GeoInterface.area`. What package can and would implement the methods?

asinghvi17: I can see GEOS warning on geodesic trait at least, and some future `S2.jl` erroring on projected trait, but that seems like something that should be done in the package, not as a GeoInterface dispatch...

evetion: > I don't really get the use-case for these traits and `GeoInterface.area`. What package can and would implement the methods?

I don't really get the question. We did discuss this extensively back at SDSL no? And there was agreement in #119 that this is necessary? I have implemented your own suggestion of an UnknownTrait() in this PR.

Essentially, we mix all kind of geometries in our ecosystem, sometimes projected, sometimes geographic. Our backing implementations (either via GEOS or GeometryOps) are correct for projected only, and will give plain wrong answers on anything geographic. So there's a need for the distinction, and a trait here seems like the logical choice. This PR is fully backwards compatible, but depwarns (so for devs only) that a distinction between project and geographic should be implemented for certain GeoInterface methods that are sensitive to it.

> But most geometry types are not going to implement these area algs right? Like GeometryOps.area already works for anything.

Except for geographic things, and that's ignoring anything already in ArchGDAL, LibGEOS and derivatives. And things like extent, which even has a fallback here (again, only valid for projected data).

> I can see GEOS warning on geodesic trait at least, and some future `S2.jl` erroring on projected trait, but that seems like something that should be done in the package, not as a GeoInterface dispatch...

This package is an interface definition, so yes by definition the package has to implement it (and in doing so determine whether it errors/warns/ignores). Any package that implements GeoInterface just has to implement crstrait(geom) = ProjectedTrait() to make the depwarn go away (which a dev only see on tests, nothing in this PR is seen by an end user).

rafaqz: But that's all orthogonal to `GeoInterface.area` needing a method for it. I'm not sure what package would add that method or why

evetion: Are we only talking about `area`, or also about `length`, `extent`, etc? The `area` was actually the example discussed in #119, here I assume we also mean the others.

If we accept we need a crstrait because the required algorithm to implement `area` is different for projected vs geographic geoms, I'd say it follows you want to dispatch on it (it's a trait after all). As this is the *interface* package, it also prescribes the method needed to implement it. Just like we prescribe that all methods do `thing(x) = thing(trait(x), x)`, and packages implement the latter, not the former, even when the package itself does not always need it.

> I'm not sure what package would add that method or why

All packages that implement GeoInterface should eventually switch from `area(::geomtrait, geom)` to `area(::crstrait, ::geomtrait, geom)`, to make explicit what kind of crstrait(s) they support for these methods. With the current PR the default is Projected, so geometries that are Geographic in nature will not work with the current `area` implementations. 

It is actually a very small (and imho elegant) change, especially if you consider the alternative that each package starts adding its own dispatch method/if/else mechanism to check the crstrait, or worse they do not do so and are silently wrong as they currently are.

I believe we require this distinction before we can even think about implementing GeoInterface for https://github.com/TLipede/S2Geometry.jl, https://github.com/wookay/H3.jl, https://github.com/anowacki/GeographicLib.jl, or the geodesic part of Proj.jl.

rafaqz: My point is most packages that own geometries don't implement `GeoInterface.area` and probably won't ever implement `area`.

Personally I mostly have GeoInterface.jl / GeoJSON.jl / Shapefile.jl or in future GeoPackage.jl objects. How and where can they dispatch to implemented `GeoInterface.area` methods?

I just can't see the practical end goal.

evetion: But that's a critique on having area in GeoInterface at all right? And more related to blessed implementations than crstraits?

rafaqz: Yes, but I would personally deprecate it rather than modifying it in this PR.

(I do actually want to hear the vision for how it will work for those types... Maybe at JuliaCon)

evetion: I can understand that, but as we don't have a blessed design yet, it's hard to deprecate anything. Perfect is the enemy of good, and I think this PR is an improvement over the current state.

And in terms of usage, I think the extent (and specifically the calc_extent fallback here), and everything that uses ArchGDAL (as it has both widely used geometries and has an area implementation) would be improved by this PR.

evetion: Ok, so I've reverted all of this for the operations, except for extent (since we have a fallback to calc_extent in there). Ideally we would move that to GeometryOps, but for now we should leave it in (would also be breaking I guess?)

src/types.jl
```diff
@@ -92,3 +92,21 @@ struct FeatureTrait <: AbstractFeatureTrait end
...
+"An AbstractCRSTrait for all coordinate reference systems"
+abstract type AbstractCRSTrait end
+"An AbstractProjectedTrait for all projected coordinate reference systems"
+abstract type AbstractProjectedTrait <: AbstractCRSTrait end
+"An AbstractGeographicTrait for all geographic coordinate reference systems"
+abstract type AbstractGeographicTrait <: AbstractCRSTrait end
+"An ProjectedTrait for all projected coordinate reference systems"
+struct ProjectedTrait <: AbstractProjectedTrait end
+"An UnknownTrait for all unknown (assumed projected) coordinate reference systems"
+struct UnknownTrait <: AbstractProjectedTrait
```
rafaqz, line 105: Should this inherit from AbstractCRStrait?

src/types.jl
```diff
@@ -92,3 +92,21 @@ struct FeatureTrait <: AbstractFeatureTrait end
...
 "A FeatureCollectionTrait holds objects of `FeatureTrait` and an `extent`"
 struct FeatureCollectionTrait <: AbstractFeatureCollectionTrait end
+
+"An AbstractCRSTrait for all coordinate reference systems"
+abstract type AbstractCRSTrait end
+"An AbstractProjectedTrait for all projected coordinate reference systems"
+abstract type AbstractProjectedTrait <: AbstractCRSTrait end
+"An AbstractGeographicTrait for all geographic coordinate reference systems"
+abstract type AbstractGeographicTrait <: AbstractCRSTrait end
+"An ProjectedTrait for all projected coordinate reference systems"
```
rafaqz, line 102:
```suggestion
"An `AbstractCRSTrait` for all geographic coordinate reference systems"
```

src/types.jl
```diff
@@ -92,3 +92,21 @@ struct FeatureTrait <: AbstractFeatureTrait end
 abstract type AbstractFeatureCollectionTrait <: AbstractTrait end
 "A FeatureCollectionTrait holds objects of `FeatureTrait` and an `extent`"
 struct FeatureCollectionTrait <: AbstractFeatureCollectionTrait end
+
+"An AbstractCRSTrait for all coordinate reference systems"
+abstract type AbstractCRSTrait end
+"An AbstractProjectedTrait for all projected coordinate reference systems"
```
rafaqz, line 98:
```suggestion
"Supertype for all coordinate reference system types"
```

src/types.jl
asinghvi17: If we assume projected anyway then dispatch can just be on `AbstractProjectedTrait` right?

src/fallbacks.jl
```diff
@@ -111,7 +111,7 @@ end
 coordinates(t::AbstractFeatureCollectionTrait, fc) = map(f -> coordinates(f), getfeature(t, fc))
 
 extent(::Any, x) = Extents.extent(x)
-function calc_extent(t::AbstractPointTrait, geom)
+function calc_extent(::AbstractProjectedTrait, t::AbstractPointTrait, geom)
```
rafaqz, line 114: Won't this break getting extents for lat/lon?

evetion: Well yes, but the extents of lat lon are wrong, as in you need to return a geographic extent to make the box useful for intersection etc?

rafaqz: If all extents are in lat/lon it's not wrong?

This would be a catastrophically breaking change

src/types.jl
evetion: Yeah, I assume you implement GeometryOps with Abstractprojected and catch both.

src/fallbacks.jl
evetion: Do antimeridian crossings work in Extents?

asinghvi17: Probably not.

rafaqz: We could do Cyclic dimensions like DimensionalData.jl has

evetion: Let's discuss further in https://github.com/rafaqz/Extents.jl/issues/27. As long as no-one implements GeographicTrait in the ecosystem, this is not breaking anything.

src/types.jl
evetion: I think you have your docstrings confused (are you on mobile?)

rafaqz: Yeah lack of lines breaks + mobile lol

src/fallbacks.jl
rafaqz: Ok lets not merge the PR until we have that worked out

rafaqz (approved):

Issue #110, opened 2023-09-04
Can't plot geometry wrapper types


evetion:
```julia
julia> using GeoInterface, Plots, GeoFormatTypes

julia> points = GeoInterface.Point.(rand(10),rand(10), crs=GeoFormatTypes.EPSG(4326))
10-element Vector{GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG}}:
 GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG}((0.6569792502348218, 0.5196554284209661), EPSG(4326))
...

julia> plot(points)
ERROR: Cannot convert GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG} to series data for plotting
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] _prepare_series_data(x::GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:8
  [3] _series_data_vector(x::GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:36
  [4] (::RecipesPipeline.var"#21#24"{Dict{Symbol, Any}})(vi::GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG})
    @ RecipesPipeline ./none:0
  [5] iterate
    @ ./generator.jl:47 [inlined]
  [6] _series_data_vector(v::Vector{GeoInterface.Wrappers.Point{false, false, Tuple{Float64, Float64}, EPSG}}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:51
  [7] macro expansion
    @ ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:129 [inlined]
  [8] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, #unused#::Type{RecipesPipeline.SliceIt}, x::Any, y::Any, z::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesBase/BRe07/src/RecipesBase.jl:300
[...]

PR #143, opened 2024-07-12
Add package extension on RecipesBase

evetion:
This moves all the plots related code, including the macro(!) into GeoInterface. This now works:

```julia
using GeoInterface, Plots
Precompiling GeoInterface
  1 dependency successfully precompiled in 1 seconds. 1 already precompiled.
[ Info: Precompiling RecipesBaseExt [b9601432-96b0-576e-9c4b-47fc12d2b5bf]
points = GeoInterface.Point.(rand(10),rand(10))
julia> plot(points)
```
Screenshot 2024-07-12 at 10 28 16

Since we now export the macros, I think the existing GeoInterfacesRecipes continues to work, even though it's now empty.


Project.toml
```diff
@@ -6,6 +6,12 @@ version = "1.3.5"
...
 
+[weakdeps]
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+
+[extensions]
+RecipesBaseExt = "RecipesBase"
+
 [compat]
 Extents = "0.1.1"
 julia = "1"
```
asinghvi17, line 17: We'd have to bump Julia compat or keep publishing GeoInterfaceRecipes to avoid breakage IMO.

```suggestion
julia = "1.9"
```

src/plots.jl
```diff
@@ -0,0 +1,151 @@
...
+```
+"""
+macro enable(typ)
+    esc(expr_enable(typ))
+end
+
+# Compat
+macro enable_geo_plots(typ)
+    esc(expr_enable(typ))
+end
```
asinghvi17, line 151: If I understand correctly, this is no longer available as a "name" to import from GeoInterface / some other package.  Do you want to keep the package GeoInterfaceRecipes and just load that code path as the extension?

asinghvi17 (changes requested):

evetion:
This at least fixes #110 (for people that forget to do `using GeoInterfaceRecipes`).

I still have to check whether this can be used to deprecate GeoInterfaceRecipes completely, and how that would work. Essentially, as soon as a package uses RecipesBase, it would trigger this package extension, but if it triggers in an extension on RecipesBase, then the macro would not be available.

src/plots.jl
evetion: The macro is exported by GeoInterface now, so it should be usable. I've answered your other question in the general comments.

src/plots.jl
```diff
@@ -0,0 +1,151 @@
+# Convert coordinates to the form used by Plots.jl
+_coordvecs(::PointTrait, geom) = [tuple(coordinates(geom)...)]
```
rafaqz, line 2: Can we put all these functions in the extension and just leave the stubs here to extend in the extension?

asinghvi17:
I'm also not sure about defining `@enable` as being only for Plots.  Since RecipesBase is pretty small, (as is MakieCore), is there an issue with keeping the packages as they are, and simply using an `include` trick on GeoInterfaceRecipes for a package extension?

evetion:
Yeah the other solution that I considered is just duplicating the code. I'm pretty unsure about include, as it's a standalone package, even though it's in a monorepo here right?

rafaqz:
Can we just do the `include("../GeoInterfaceRecipes/src/GeoInterfaceRecipes.jl") and `include("../GeoInterfaceMakie/src/GoInterfaceMakie.jl") tricks in extensions GeoInterfaceRecipesExt and GeoInterfaceMakieExt?

Then we don't have the circular dep on the sub-packages but we can use the same code but not break anything.

It does end up using a bit more compilation and memory as its the same code used twice, but I think that's ok.

asinghvi17:
GIMakieExt is circular but RecipesExt could work

rafaqz:
Oh of course because Makie depends directly on GeoInterface.jl via GeometryOps.jl, forgot for a second.

Maybe... could we just move the GeometryOps geointerface code to an extension in GeoInterface.jl?

We need to solve this somehow

asinghvi17:
I think we tried to in a GeometryBasics PR but I ended up not wanting to do that for some reason.  We can in theory just have Makie depend on GeoInterfaceMakie too...

evetion:
> Can we just do the `include("../GeoInterfaceRecipes/src/GeoInterfaceRecipes.jl")` and `include("../GeoInterfaceMakie/src/GoInterfaceMakie.jl")` tricks in extensions GeoInterfaceRecipesExt and GeoInterfaceMakieExt?
> 
> Then we don't have the circular dep on the sub-packages but we can use the same code and not break anything.
> 
> It does end up using a bit more compilation and memory as its the same code used twice, but I think that's ok.
> 
> Edit: we may need to move code out of the modules into separate files for this to work.

While it seems to work locally, I at least get cache misses for precompiling. I don't think it's intended for code to be included outside of the ext folder/file.

```[ Info: Precompiling RecipesBaseExt [b9601432-96b0-576e-9c4b-47fc12d2b5bf] (cache misses: include_dependency fsize change (2))```

**Comment by evetion** [comment]
Ok, moved all the conversion code to the extension so it doesn't slow down GeoInterface itself by default. Only the macro stubs are now left in.

GeoInterfaceRecipes is now empty, it imports RecipesBase and GeoInterface and only re-exports the enable macros. It depends on Julia 1.9 and this version of GeoInterface, so if someone loads this new version it will trigger the package extension. If you have an older version, even though the package extension will trigger, you should get the macro from GeoInterfaceRecipes itself. This effectively deprecates GeoInterfaceRecipes.

I will test this with ArchGDAL (it depends directly on GeoInterfaceRecipes).

**Comment by rafaqz** [comment]
Cool! Now we need to figure out how to do this with Makie. If we move the GeoInterface code in GeometryBasics to an extension here instead that should work too?

There will be some juggling to do to not break things in the handover

**Comment by evetion** [comment]
> Cool! Now we need to figure out how to do this with Makie. If we move the GeoInterface code in GeometryBasics to an extension here instead that should work too?
> 
> There will be some juggling to do to not break things in the handover

It should work, although I'm not that familiar with the triangle of Makie(Core?)/GeometryBasics/GeoInterface. But thinking about it, it might clash on the @enable macro if you also load RecipesBase, since we moved the stub to the shared GeoInterface here.

**Comment by rafaqz** [comment]
Oh no the `@enable` macro. Can we rename the verion in this PR? It shouldn't be `@enable` in GeoInterface anyway because it unclear what that is, where `GeoInterfaceRecipes.@enable` was clear

Probably we need to define e.g. `@enable_plots` here and call the same internal code as `@enable`, rather than actually moving `@enable` here.

**Comment by evetion** [comment]
> Oh no the `@enable` macro. Can we rename it in this PR? It shouldn't be `@enable` in GeoInterface anyway because it unclear what that is, where `GeoInterfaceRecipes.@enable` was clear
> 
> Probably we need to define e.g. `@enable_plots` here and call the same internal code as `@enable`, rather than actually moving `@enable` here.

We need two, or weird stuff with Module dispatch in the macro (1.10+). So I would go with
- `@enable_plots`
- `@enable_makie`

File: src/plots.jl [hunk 4]
```diff
@@ -0,0 +1,3 @@
+# Stubs are extended in ext/RecipesBaseExt.jl
+macro enable() end
```
rafaqz, line 2: This means packages need to do `import RecipesBase` before using this macro? Is that the idea?

evetion: That would work, but like my example, I would expect `using Plots`. Anything that would load RecipesBase will trigger it.

evetion: And in terms of packages, yes RecipesBase would work, and so would their existing dependency on GeoInterfaceRecipes (since it depends on RecipesBase).

rafaqz:
Yeah thats what I mean `@enable_plots` now, and `@enable_makie` later

evetion:
With the latest commits, this now works:

```julia
using Plots
using GeoInterface

GeoInterface.geoplot(rand(Tuple{Float64, Float64}, 100))
```
Screenshot 2024-10-13 at 17 50 07

Using the extension, we can plot any GeoInterface compatible geometry/feature/collection. It required some more hacking of RecipesBase, mostly macroexpanding `@userplot` (for `geoplot`) and `@recipe` (for collections).

This should be fully backwards compatible with GeoInterfaceRecipes. I've anticipated a Makie(Core) extension, so `geoplot` could work with Makie in the future as well. The only problem is when both extensions are loaded, we'll need some fix for that in a future PR.

asinghvi17:
What's the motivation for `geoplot`?  I thought `plot` from Plots worked fine for all featurecollections / geometries?

rafaqz:
Yes isn't the idea that `plot` just works?

evetion:
`plot` works if a package also depends on GeoInterfaceRecipes/RecipesBase and does `@enable_plot`. `geoplot` works for all GeoInterface enabled geometries.

asinghvi17:
I get the feeling that the number of geometries that are GeoInterface enabled but don't have `@enable` macros in extensions has got to be pretty low by now...also, shouldn't it be the solution to tell people to use `@enable` instead of a new function they'll have to learn?  Enable is generic across all plot types so trying `lines` or `scatter` will work equally well - it just gives more options.

Now that the macros will actually be accessible it seems reasonable to do, at least to me...

rafaqz:
I was also hoping `plot` would work universally, it almost does already

evetion:
Ah fair enough, my two major gripes were
- forgetting to do use GeoInterfaceRecipes as user
- forgetting to depend on GeoInterfaceRecipes as producer 

The latter is indeed rarer, and not *that* important. And geoplot is a new thing indeed, so let's say that this was just a nice exercise in hacking RecipesBase.

I'll revert the `geoplot` thing, and then this should be good to go.

asinghvi17:
> forgetting to depend on GeoInterfaceRecipes as producer

That will probably continue to be an issue AFAICT, we can document it better though.

> forgetting to do use GeoInterfaceRecipes as user

My understanding is that this PR makes it so that this should not be an issue?

GeoInterfaceRecipes/Project.toml
```diff
@@ -1,16 +1,16 @@
...
-version = "1.0.2"
+version = "1.0.3"
 
 [deps]
 GeoInterface = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 
 [compat]
-GeoInterface = "1"
+GeoInterface = "1.3.7"
```
rafaqz, line 11:
```suggestion
GeoInterface = "1.3.8"
```

evetion:
This (slightly outdated) PR seems relevant to the discussion in https://github.com/JuliaGeo/Shapefile.jl/issues/138. Although it still requires the package to have a weakdep, because they own the geometry types.

evetion:
Closed in favor of #198

Issue #126, opened 2024-04-14
Fix printing for wrapper geoms


asinghvi17:
The wrapper geoms dump all info to the REPL when printed, which is quite annoying.  A Tables.jl like overview would be cool for feature collections, and perhaps something like Shapefile.jl printing for wrapper geoms?  We could even mock up the WKT.

rafaqz:
Sounds great

asinghvi17:
Sweet.  I'll get to this at some point in the next few weeks unless someone else does it first :D

asinghvi17:
@bencurran98 worked on this at the hackathon, FYI so people don't pick this up and duplicate work!

PR #150, opened 2024-07-25
Fix printing wrapper geoms

BenCurran98:
Proposing a fix for [this issue](https://github.com/JuliaGeo/GeoInterface.jl/issues/126) where we modify the displayed text for our wrapper geometries. Each geometry can be displayed in "regular" or "compact" form which will influence how much information is shown to the user. This should make things a bit clearer so we don't get spammed with the raw fields that can be hard to follow.

### DIFF
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -88,7 +88,7 @@ function ngeom(trait::AbstractGeometryTrait, geom::WrapperGeometry{<:Any,<:Any,T
     isgeometry(T) ? ngeom(parent(geom)) : length(parent(geom))
 end
 
-# We eefine all the types in a loop so we have standardised docs and behaviour
+# We define all the types in a loop so we have standardised docs and behaviour
 # without too much repetition of code.
 # `child_trait` and `child_type` define the trait and type of child geometries
 # a geometry can be constructed from.
@@ -171,6 +171,40 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
         convert(::Type{$geomtype}, ::$trait, geom) = $geomtype(geom)
         # But not if geom is already a WrapperGeometry
         convert(::Type{$geomtype}, ::$trait, geom::$geomtype) = geom
+        
+        function Base.show(io::IO, ::MIME"text/plain", geom::$geomtype{Z, M, T, E, C}; show_mz::Bool = true) where {Z, M, T, E <: Union{Nothing,Extents.Extent}, C}
+            compact = get(io, :compact, false)
+            print(io, $geomtype)
+            if show_mz
+                print(io, "{$Z, $M}")
+            end
+            print(io, "(")
+            this_geom = getgeom(trait(geom), geom)
+            if this_geom isa AbstractVector
+                print(io, "[")
+                for (i, g) ∈ enumerate(this_geom)
+                    _nice_print_geom(io, g, false)
+                    if i != length(this_geom)
+                        print(io, ",$(compact ? "" : " ")")
+                    end
+                end
+                print(io, "]")
+            else
+                show(io, g; show_mz = false)
+            end
+            if compact
+                print(io, ")")
+            else
+                if !isnothing(geom.extent)
+                    print(", extent = $(geom.extent)")
+                end
+                if !isnothing(geom.crs)
+                    print(", crs = $(geom.crs)")
+                end
+                print(")")
+            end
+            return nothing
+        end
     end
     @eval function $geomtype{Z,M}(geom::T; extent::E=nothing, crs::C=nothing) where {Z,M,T,E,C}
         Z isa Union{Bool,Nothing} || throw(ArgumentError("Z Parameter must be `true`, `false` or `nothing`"))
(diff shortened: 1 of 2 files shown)


asinghvi17:
This looks great, thanks!  Love the compactification of the elements you have going on there. 

One more thing that I'm thinking we should do is to decrease the number of points printed, so that that any geometry fits on one line.  Otherwise you get crazy long strings, aka what we have now, which have crashed my terminal a couple times when ultra long geometries are considered (10k points or higher).

Consider something like the following:

```julia
screen_nrows, screen_ncols = displaysize(io)
# limit the number of points plotted to 
# floor((screen_ncols - currently_used_space - 1 #= triple dot character =#) / length_of_one_point_in_chars)
# add a `...` in between (unicode character) or indicate number of points somehow, as is done in Strings
```

I will post a more useful prototype sometime later today, but this is the basic concept.

BenCurran98:
@asinghvi17 I've had a crack at this, is this sort of what you're after? Now we compress geometries with "..." and report the number of sub geometries hidden, e.g.
```julia
linestring = GI.LineString(map(_ -> rand(2), 1:10_000))
```
Gives
```julia
LineString{false, false}([[0.6022716265309537, 0.46854202492808894], [0.7698994281922478, 0.5141934816807233],  … (9997) … [0.5275211157024845, 0.7022242690985355], [0.09265052455532974, 0.5388384079940159]])
```

asinghvi17:
Sorry for the delay!  This is next on my list.  Should get to it by Wednesday.

rafaqz:
Where is this one at? would be nice to merge

asinghvi17:
I'm using it locally but have a few fixes I need to push.  Other than that it's great!

BenCurran98:
Hey @asinghvi17 , sorry for the delay (been super swamped at work lately)
I've updated some of the logic here to be a bit nicer with how we cut down the displayed details with nested geom types (like polygons with linear rings, etc.) so now we cut a bit out of each sub geometry recursively, e.g.

```julia
linearring = GI.LinearRing([(1, 2), (3, 4), (5, 6), (1, 2)])
polygon = GI.Polygon([linearring, linearring])

# gives "Polygon{false, false}([LinearRing([(1, 2), … (2) … , (1, 2)]), LinearRing([(1, 2), … (2) … , (1, 2)])])"
```

src/wrappers.jl
```diff
@@ -228,6 +301,45 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
     end
 end
 
+function _nice_geom_str(geom, ::Bool, ::Bool, ::Int) 
+    io = IOBuffer()
+    show(io, MIME("text/plain"), geom)
+    return String(take!(io))
+end
+
```
rafaqz, line 309:
```suggestion
```

src/wrappers.jl
```diff
@@ -228,6 +301,45 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
...
+# handle tuples/vectors explicitly
+function _nice_geom_str(geom::AbstractVector, ::Bool, compact::Bool, ::Int)
+    spacing = compact ? "" : " "
+    str = "["
+    str *= _add_elements_with_spacing(geom, spacing)
+    str *= "]"
+    return str
+end
+
```
rafaqz, line 326:
```suggestion
```

src/wrappers.jl
```diff
@@ -228,6 +301,45 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
...
+# need a work around to pass the show_mz variable through - put string to a temp IOBuffer then read it
+function _nice_geom_str(geom::WrapperGeometry, show_mz::Bool, compact::Bool, screen_ncols::Int) 
+    buf = IOBuffer()
+    io = IOContext(IOContext(buf, :compact => compact))
+    show(io, MIME("text/plain"), geom; show_mz = show_mz, screen_ncols = screen_ncols)
+    return String(take!(buf))
+end
+
```
rafaqz, line 317:
```suggestion
```

src/wrappers.jl
```diff
@@ -377,6 +519,36 @@ function Feature(geometry=nothing; properties=nothing, crs=nothing, extent=nothi
     end
 end
 
+function Base.show(io::IO, ::MIME"text/plain", f::Feature; show_mz::Bool = true)
+    compact = get(io, :compact, false)
+    spacing = compact ? "" : " "
+    print(io, "Feature(")
+    Base.show(io, MIME("text/plain"), f.parent.geometry; show_mz = show_mz)
```
rafaqz, line 526:
```suggestion
    show(io, MIME("text/plain"), f.parent.geometry; show_mz = show_mz)
```

src/wrappers.jl
```diff
@@ -377,6 +519,36 @@ function Feature(geometry=nothing; properties=nothing, crs=nothing, extent=nothi
...
+    non_geom_props = filter(!=(:geometry), propertynames(f.parent))
+    if !isempty(non_geom_props)
+        print(io, ",$(spacing)properties$(spacing)=$(spacing)(")
+        for (i, property) ∈ enumerate(non_geom_props)
+            print(io, "$(property)$(spacing)=$(spacing)")
+            Base.show(io, getproperty(f.parent, property))
```
rafaqz, line 532:
```suggestion
            show(io, getproperty(f.parent, property))
```

rafaqz (approved):
Looks good to me, lets merge. 

Comments are just simple code cleanup

PR #152, opened 2024-08-08
[WIP] Add a fallback to `geometry` for FeatureCollections

asinghvi17:
What do people think of this?

This enables:
```julia
tab = Shapefile.Table(...)
shapes = GI.geometry(tab)

# or, and more importantly,

tab = GeoJSON.read(...)
shapes = GI.geometry(tab)
```

This can always be customized for individual tables for efficiency's sake, but from a UI perspective I think this is a solid improvement over what we have now.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -52,7 +52,9 @@ geometrycolumns(featurecollection) = (:geometry,)
 Retrieve the geometry of `feat`. It is expected that `isgeometry(geom) === true`.
 Ensures backwards compatibility with GeoInterface version 0.
 """
-geometry(feat) = nothing
+geometry(feat) = geometry(trait(feat), feat)
+geometry(::Union{AbstractTrait, Nothing), feat) = nothing
+geometry(::FeatureCollectionTrait, fc) = Iterators.map(geometry, getfeature(fc))
 
 """
     GeoInterface.properties(feat) => properties


rafaqz:
Seems sensible to me? Maybe a little confusing that its not `getgeom` and also allows `getgeom(fc, i)` ?

asinghvi17:
`getgeom` doesn't work for Features by definition though, we may want to change the API spec in some theoretical 2.0 to make that work?  I guess the idea back then was probably a separation of interests between Features and Geometries (see e.g `trait` vs `geomtrait`)...

rafaqz:
I'm not sure what you mean "by definition"... we are talking about `FeatureCollection` not `Feature` and its currently just undefined.

We could equally say `geometry` doesn't work for FeatureCollection "by definition".

The reason to use `getgeom` is that it always works on a list of geometries, and can choose an indexed geometry, while `geometry` currently just returns a single geometry.

rafaqz:
Just hit some code needing this pretty badly... 

But I do think `getgeom` is a more suitable method for getting an iterable of geometries or a single geometry by index from from a feature collection - as it matches the same behavior and return values as for a geometry collection. 

@evetion any thoughts on this?

evetion:
So we shot ourselves in the foot with the `geometry` name for Features (to be compatible with pre v1 of GeoInterface). The Feature(Collection) part of the spec has always a been underdeveloped.

## Current
`properties(::Feature)`
`geometry(::Feature)`

`getfeature(::FeatureCollection)` and `getfeature(::FeatureCollection, i)`
`geometrycolumns(::FeatureCollection) = (:geometry,)` can be multiple!

## Suggestion
- deprecate `geometry` in favor of `getgeom` (just rename) and think about `getprop`?
- `getgeom` on fc is just a `getgeom(getfeature)` iterator
- think well about handling multiple geometries using `geometrycolumns`. I think the spec forbids a feature from having multiple geometries, but given we equate a row in a table with a feature, we implicitly allow it.

PR #154, opened 2024-08-27
Better child error when creating geom from array

tiemvanderdeure:
When creating a geometry from an array that contains elements with the wrong trait, refer to the first element that has the wrong trait, rather than the first element of the array.

MWE:
```julia
import GeoInterface as GI
p = GI.Point(0,0)
x = 1
GI.Line([p, x])
```

Before this PR throws a

```
ERROR: ArgumentError: GeoInterface.LineTrait must have child objects with trait GeoInterface.PointTrait, got GeoInterface.Wrappers.Point{false, false, Tuple{Int64, Int64}, Nothing} with trait 
GeoInterface.PointTrait()
```

After this PR throws a: 
```
ERROR: ArgumentError: LineTrait must have child objects with trait PointTrait, got Int64 with trait nothing
```

### DIFF
--- a/src/wrappers.jl
+++ b/src/wrappers.jl
@@ -186,10 +186,10 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
         # Otherwise wrap an array of child geometries
         elseif geom isa AbstractArray
             child = first(geom)
-            chilren_match = all(child -> geomtrait(child) isa $child_trait, geom)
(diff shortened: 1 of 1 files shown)


tiemvanderdeure:
@rafaqz @asinghvi17 could one of you review/merge this?

rafaqz (approved):

PR #156, opened 2024-09-20
[WIP] Add a Tables.jl extension for FeatureCollection and Feature

asinghvi17:
This PR treats FeatureCollections as row tables, and Features as rows.  

```julia
julia> @time DataFrame(fc2)
  6.790032 seconds (10.19 M allocations: 3.447 GiB, 1.13% gc time)
177×169 DataFrame
 Row │ geometry                           FCLASS_IL        CONTINENT      NAME_FA              WB_A3   ADM0_A3_WB  FCLASS_SE        NAME_RU                ⋯
     │ WrapperG…                          Union…           String         String               String  Int64       Union…           String                 ⋯
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ MultiPolygon{false, false, Vecto…                   Oceania        فیجی                 FJI            -99                   Фиджи                  ⋯
   2 │ Polygon{false, false, Vector{Lin…                   Africa         تانزانیا             TZA            -99                   Танзания
   3 │ Polygon{false, false, Vector{Lin…                   Africa         صحرای غربی           -99            -99                   Западная Сахара
   4 │ MultiPolygon{false, false, Vecto…                   North America  کانادا               CAN            -99                   Канада
   5 │ MultiPolygon{false, false, Vecto…                   North America  ایالات متحده آمریکا  USA            -99                   США                    ⋯
  ⋮  │                 ⋮                         ⋮               ⋮                 ⋮             ⋮         ⋮              ⋮                  ⋮             ⋱
 173 │ Polygon{false, false, Vector{Lin…                   Europe         صربستان              SRB            -99                   Сербия
 174 │ Polygon{false, false, Vector{Lin…                   Europe         مونتهنگرو            MNE            -99                   Черногория
 175 │ Polygon{false, false, Vector{Lin…  Admin-0 country  Europe         کوزووو               KSV            -99  Admin-0 country  Республика Косово
 176 │ Polygon{false, false, Vector{Lin…                   North America  ترینیداد و توباگو    TTO            -99                   Тринидад и Тобаго      ⋯
 177 │ Polygon{false, false, Vector{Lin…  
```
so it's not great, but we can make it work.  I'm still not sure how GeoJSON gets it so fast.

### DIFF
--- a/ext/GeoInterfaceTablesExt.jl
+++ b/ext/GeoInterfaceTablesExt.jl
@@ -1,0 +1,126 @@
+module GeoInterfaceTablesExt
+
+using GeoInterface
+using GeoInterface.Wrappers
+using Tables
+
+# This module is meant to extend the Tables.jl interface to features and feature collections, such that they can be used with Tables.jl.
+# This enables the use of the Tables.jl ecosystem with GeoInterface wrapper geometries.
+
+# First, define the Tables interface
+
+Tables.istable(::Type{<: Wrappers.FeatureCollection}) = true
+Tables.isrowtable(::Type{<: Wrappers.FeatureCollection}) = true
+Tables.rowaccess(::Type{<: Wrappers.FeatureCollection}) = true
+Tables.rows(fc::Wrappers.FeatureCollection{P, C, E}) where {P <: Union{AbstractArray{<: Wrappers.Feature}, Tuple{Vararg{<: Wrappers.Feature}}}, C, E} = GeoInterface.getfeature(fc)
+Tables.rows(fc::Wrappers.FeatureCollection) = Iterators.map(Wrappers.Feature, GeoInterface.getfeature(fc))
+Tables.schema(fc::Wrappers.FeatureCollection) = nothing
+
+# Define the row access interface for feature wrappers
+function Tables.getcolumn(row::Wrappers.Feature, i::Int)
+    if i == 1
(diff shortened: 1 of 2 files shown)


ext/GeoInterfaceTablesExt.jl
```diff
@@ -0,0 +1,129 @@
...
+# Credit to [Rafael Schouten](@rafaqz)
+# Adapted from JSONTables.jl jsontable method
+# We cannot simply use their method as we have concrete types and need the key/value pairs
+# of the properties field, rather than the main object
+# TODO: Is `missT` required?
+# TODO: The `getfield` is probably required once
+missT(::Type{Nothing}) = Missing
+missT(::Type{T}) where {T} = T
+
+function property_schema(features)
```
rafaqz, line 40: Will this work for GeoJSON too? Would be good to delete the code there if it does

asinghvi17: It would but better not to rely on getting it out of the extension

rafaqz: I mean if we want to use it in GeoJSON it needs to be an interface method. But yeah lets not bother.

rafaqz:
Lots of licked cooky PRs here unfinished ;)

Would be good to get all of these in

PR #160, opened 2024-09-21
Support tuples of child geoms directly, wrap single child geom in tuple

asinghvi17:


rafaqz:
To avoid the allocation?

asinghvi17:
Yeah, just to make it simple.  This way I can use this in GeometryOps on all the geos methods too, and I don't have to call tuples.

rafaqz:
It's slightly concerning for type stability. Now vectors of these geoms have length in the types

src/wrappers.jl
```diff
@@ -180,8 +180,8 @@ for (geomtype, trait, childtype, child_trait, length_check, nesting) in (
         # Is geom a single geometry ?
         if isgeometry(geom)
             if geomtrait(geom) isa $child_trait
-                # If geom is a child_trait, then make geom a vector and call again
-                return $geomtype([geom]; extent, crs)
+                # If geom is a child_trait, then make geom a tuple and call again
+                return $geomtype((geom,); extent, crs)
```
asinghvi17, line 184: If it's bad for type stability, we can always revert this.  Users can choose to simply write `MultiType((geom,))` if it suits them.

alex-s-gardner (commented):
looks good to me

alex-s-gardner:
@rafaqz Is this getting rejected?

rafaqz:
Quite possibly ;)

asinghvi17:
Let's make this a different PR then, I'm happy with the first one

rafaqz:
Probably people can just use Static arrays if they need this? It's more robust as the non mutating vector methods will still work

rafaqz:
Lets close this, I think StaticArrays is better

PR #161, opened 2024-09-21
Implement crs and geometrycolumn fallbacks via DataAPI

asinghvi17:
- [x] Needs tests
- [x] needs docs

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -95,8 +95,11 @@ issimple(t::AbstractMultiPointTrait, geom) = allunique((getgeom(t, geom)))
 issimple(t::AbstractMultiCurveTrait, geom) = all(issimple.(getgeom(t, geom)))
 isclosed(t::AbstractMultiCurveTrait, geom) = all(isclosed.(getgeom(t, geom)))
 
-crs(::Nothing, geom) = _get_dataapi_metadata(geom, "GEOINTERFACE:crs", nothing) # see `metadata.jl`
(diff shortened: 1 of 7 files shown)


src/fallbacks.jl
```diff
@@ -95,8 +95,17 @@ issimple(t::AbstractMultiPointTrait, geom) = allunique((getgeom(t, geom)))
...
+crs(::AbstractTrait, geom) = _get_dataapi_metadata(geom, "GEOINTERFACE:crs", nothing)
+
+function _get_dataapi_metadata(geom::GeomType, key, default) where GeomType
+    if DataAPI.metadatasupport(GeomType).read
+        if key in DataAPI.metadatakeys(geom)
+            return DataAPI.metadata(geom, key; style = false)
+        end
+    end
+    return default
+end
```
asinghvi17, line 108: Should I move this to some other location / file?

evetion: Maybe to (a new) metadata.jl? I discussed having a Geometadata.jl package with @rafaqz, but that's probably overkill for now.

rafaqz:
You mean if its a single `String` we should return ` (Symbol(str),)`

asinghvi17:
Yes

rafaqz:
Sounds pragmatic

asinghvi17:
OK, I've added some aftercare functions here.  Do we want anything else?

rafaqz (approved):
LGTM. Can we merge?

PR #166, opened 2024-09-25
skip nan in extent calculation

rafaqz:
This PR fixes the problem where NaNs win over other coordinates to make `NaN` extents that don't work anywhere. Or do work in DimensionalData.jl and because `NaN` comparisons always return `false`, it just select the maximum possible extent.

To do this I switch `extrema` to `_nan_free_extrema`, which turns out to be faster anyway.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -123,10 +123,10 @@ function calc_extent(t::AbstractPointTrait, geom)
 end
 function calc_extent(t::AbstractGeometryTrait, geom)
     points = getpoint(t, geom)
-    X = extrema(p -> x(p), points)
-    Y = extrema(p -> y(p), points)
+    X = _nan_free_extrema(x, points)
+    Y = _nan_free_extrema(y, points)
     if is3d(geom)
-        Z = extrema(p -> z(p), points)
+        Z = _nan_free_extrema(z, points)
         Extent(; X, Y, Z)
     else
         Extent(; X, Y)
@@ -138,6 +138,24 @@ function calc_extent(::AbstractFeatureTrait, feature)
     isnothing(geom) ? nothing : extent(geom)
 end
 calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
+
+function _nan_free_extrema(f, points)
+    agg_min = agg_max = f(first(points))
+    found = false
+    for point in points
+        c = f(point)
+        isnan(c) && continue
+        if found
+            agg_min = min(agg_min, c)
+            agg_max = max(agg_max, c)
+        else
+            found = true
+            agg_min = c
+            agg_max = c
+        end
+    end
+    return agg_min, agg_max
+end
 
 # Package level `GeoInterface.convert` method
 # Packages must implement their own `traittype` method
--- a/test/test_wrappers.jl
+++ b/test/test_wrappers.jl
@@ -301,6 +301,13 @@ fc = GI.FeatureCollection(fc_unwrapped.parent; crs=EPSG(4326), extent=GI.extent(
 vecfc = GI.FeatureCollection([(geometry=(1,2), a=1, b=2)])
 @test GI.getfeature(vecfc, 1) == (geometry=(1,2), a=1, b=2)
 
+@testset "Extent skips NaN" begin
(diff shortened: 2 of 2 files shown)


src/fallbacks.jl
```diff
@@ -139,6 +139,24 @@ function calc_extent(::AbstractFeatureTrait, feature)
 end
 calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filter(!isnothing, collect(extent(f) for f in getfeature(t, fc))))
 
+function _nan_free_extrema(f, points)
+    agg_min = agg_max = f(first(points))
```
asinghvi17, line 143: This is vulnerable to `first(points) == (NaN, NaN)` but I don't imagine that's a scenario that will happen often.  Even so, maybe a check could be good?

asinghvi17: Ah I guess the tests proved me wrong there? But I'm not sure how that works here...

rafaqz: We loop over all the points and the found switch means we use the first non-NaN. These are only used if we never find anything else, in which case NaN is the best we have.

asinghvi17: Aha, I missed that.  Looks good to me then!

test/test_wrappers.jl
```diff
@@ -301,6 +301,13 @@ fc = GI.FeatureCollection(fc_unwrapped.parent; crs=EPSG(4326), extent=GI.extent(
 vecfc = GI.FeatureCollection([(geometry=(1,2), a=1, b=2)])
 @test GI.getfeature(vecfc, 1) == (geometry=(1,2), a=1, b=2)
 
+@testset "Extent skips NaN" begin
+    nan_multipoint = GI.MultiPoint([(1.0, NaN), (3.0, 4.0), (3.0, 2.0), (NaN, 4.0), (7.0, 8.0), (9.0, 10.0)])
+    nan_polygon = GI.Polygon(GI.LineString([(1.0, NaN), (3.0, 4.0), (3.0, 2.0), (NaN, 4.0), (7.0, 8.0), (9.0, 10.0)]))
+    @test GI.extent(nan_multipoint) == Extent(X = (1.0, 9.0), Y = (2.0, 10.0))
+    @test GI.extent(nan_polygon) == Extent(X = (1.0, 9.0), Y = (2.0, 10.0))
+end
```
asinghvi17, line 309:
```suggestion
@testset "Extent skips NaN" begin
    nan_multipoint = GI.MultiPoint([(1.0, NaN), (3.0, 4.0), (3.0, 2.0), (NaN, 4.0), (7.0, 8.0), (9.0, 10.0)])
    nan_polygon = GI.Polygon(GI.LineString([(1.0, NaN), (3.0, 4.0), (3.0, 2.0), (NaN, 4.0), (7.0, 8.0), (9.0, 10.0)]))
    @test GI.extent(nan_multipoint) == Extent(X = (1.0, 9.0), Y = (2.0, 10.0))
    @test GI.extent(nan_polygon) == Extent(X = (1.0, 9.0), Y = (2.0, 10.0))

    all_nan_multipoint = GI.MultiPoint([(NaN, NaN) for i in 1:10])
    all_nan_ext = GI.extent(all_nan_multipoint)
    @test all(isnan, all_nan_ext.X)
    @test all(isnan, all_nan_ext.Y)
end

asinghvi17 (approved):

test/test_wrappers.jl
asinghvi17: Just adds a test to make sure that all-NaN vectors keep returning NaNs.  It's a bit convoluted since `NaN != NaN`...

rafaqz:
The weird thing about this (that I just realised for Extents.jl as well) is that for `(X=NaN, Y=1000.0)` the `1000.0` will still effect the final extent. Maybe this is an extreme edge case and doesn't matter much, I'm not sure.

rafaqz:
@evetion would be good to get a review and merge on this

evetion:
LGTM, I haven't hit this specific use case yet. Can you document the behaviour in the docstring(test?) somewhere. All NaN points will still yield a NaN extent right?

rafaqz:
Fir sure, a doc comment would help. And yes all NaN still returns NaN

asinghvi17 (approved):
LGTM, I added some documentation as well.

PR #167, opened 2024-09-25
add `coordtype`

### LINKED ISSUE #128: Add a `coordtype` method
I'm not sure if this is possible (do any formats allow mixed coordinate types like Float32 and Float64 in the same dataset on even the same polygon??)

But it would be good to be able to get the coord type at any level, e.g. on a feature or polygon as well as a point. Then we can pass it into loops to fix type stability, and use it to e.g. preallocate vectors.

@asinghvi17 @skygering were discussing this on slack recently. Currently in GeometryOps,jl we just default to Float64 and allow user override. But it could be nice to detect the coord type and use that.

rafaqz:
Closes #128 

Currently on a GI.MultiPolygon this takes 4ns. Its not a compile time operation. We could put this in the type parameters of the wrapper geometries at construction so its instant.

A lot of packages will be able to set this absolutely to a single type, e.g. Shapefile.jl, GeoJSON.jl and LibGEOS.jl can return `Float64`.


src/wrappers.jl
```diff
@@ -318,6 +318,11 @@ convert(::Type{Point}, ::PointTrait, geom) = Point(geom)
 convert(::Type{Point}, ::PointTrait, geom::Point) = geom
 extent(trait::PointTrait, geom::Point) = extent(trait, parent(geom)) 
 
+coordtype(::Point{<:Any,<:Any,<:AbstractArray{T}}) where T = T 
+coordtype(::Point{<:Any,<:Any,<:NTuple{<:Any,T}}) where T = T 
+coordtype(::Point{<:Any,<:Any,<:NamedTuple{<:Any,<:NTuple{<:Any,T}}}) where T = T
```
asinghvi17, line 323: Should we remove the `Point` from here, so that the dispatch with `parent(p)` for `Point` feeds into these?

test/test_wrappers.jl
asinghvi17, line 1: should we also add some `@test_inferred`?  Or is it too early for that?

rafaqz: Yes, good idea

src/wrappers.jl
rafaqz: I think that will hit the fallback method in fallbacks.jl. It just get the X coord type

asinghvi17:
GeoInterface implementations:

Bridge packages

- [ ] ArchGDAL
- [ ] LibGEOS

Readers

- [ ] GeoParquet
- [ ] GeoArrow
- [ ] GeoJSON
- [ ] Shapefile
- [ ] WellKnownGeometry

Other implementations
- [ ] GeometryBasics

evetion:
I'm fine with this, but might be good to give a concrete example of what this improves. Do we also want to give some hints using coordtype for let's say GI.coordinates? That's the most type unstable thing I can think of in here.

rafaqz:
Yeah, we could use it for `coordinates` in some cases

asinghvi17:
https://gdal.org/en/stable/doxygen/classOGRPoint.html#a5170ea70ce7458059e4395f852fce687 indicates that ArchGDAL must also set this to Float64, so it's just Julia types that need to have the type statically encoded somehow.

asinghvi17:
bump @evetion @visr - can we get this in?  I do have quite a few uses for it in GeometryOps now...

evetion:
> bump @evetion @visr - can we get this in? I do have quite a few uses for it in GeometryOps now...

Let's get it in.

PR #169, opened 2024-10-05
Correctly implement interface for wrapper geometry.

evetion:
Fixes https://github.com/evetion/WellKnownGeometry.jl/issues/37.

Packages should implement the two argument interface methods for geometries, and #124 did not. In WKG I assumed I could call `ncoord(trait, geom)`, which failed.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -12,16 +12,17 @@ function testgeometry(geom)
 
     if type == PointTrait()
         n = ncoord(geom)
+        @assert n == ncoord(type, geom) "$geom does not correctly implement three argument `getcoord`"
         if n >= 1  # point could be empty
-            getcoord(geom, 1)  # point always needs at least 2
+            c = getcoord(geom, 1)  # point always needs at least 2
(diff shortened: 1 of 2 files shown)


rafaqz:
Beat me to it! Tests?

evetion:
> Beat me to it! Tests?

Yeah, precisely what? Hasmethod for the two argument function and implement that in testgeometry? That way we can re-use for other packages who implementend it incorrectly (might be breaking some tests somewhere else).

rafaqz:
I mean just calling those methods with the trait somewhere or other, but certainly add to the generic tests if that makes more sense, we don't do that enough...

evetion:
> I mean just calling those methods with the trait somewhere or other, but certainly add to the generic tests if that makes more sense, we don't do that enough...

I've added tests (which failed on master) to the generic test implementation :)

rafaqz:
Would we not just call the methods?

evetion:
> Would we not just call the methods?

Ok done!

rafaqz (approved):

rafaqz:
Docs test fails look like Julia 1.11 GitHub actions problems, let's merge

PR #170, opened 2024-10-06
Update test_wrappers.jl to use testgeom

asinghvi17:


evetion:
Good call, but I'm sorry to say that (most? of) these are duplicates (see line 17 for example).

rafaqz:
Yeah, the real problem is `testgeometry` doesn't test anywhere near enough things

asinghvi17:
Ah, I guess I missed reading that :D - will see if there are any actually missing cases and add those.

PR #173, opened 2024-10-23
Remove ambiguity and implement `getgeom`, `ngeom` for `Wrappers.Point`

asinghvi17:
This was mentioned on Slack, the output is probably a bit better than a method ambiguity error now.


rafaqz:
Not sure `nothing` is the best return value here.

We could also just return the point and ngeoms could be 1

asinghvi17:
Error might be best?

rafaqz:
Error is pretty safe as we can always change it later, but returning `nothing` is permanent

rafaqz:
Maybe `GI.getgeom(point, 1)` should return the point? a bit like how numbers work just returning themselves?

asinghvi17:
Eh that sounds too recursive for my liking...let's just error

PR #179, opened 2025-01-13
Test isgeometry on type of geometry.

evetion:
The interface actually tells you to do so, but it's easy to forget/skip. Found this for `isgeometry(GFT.WellKnownBinary)` (blessed piracy in [WellKnownGeometry](https://github.com/evetion/WellKnownGeometry.jl/pull/41)) when working on [GeoArrow](https://github.com/JuliaGeo/GeoArrow.jl/pull/2) purely based on the types of a column (not the first instance or similar).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "GeoInterface"
 uuid = "cf35fbd7-0cd7-5166-be24-54bfbe79505f"
 authors = ["JuliaGeo and contributors"]
-version = "1.4.0"
+version = "1.4.1"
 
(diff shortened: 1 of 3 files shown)


rafaqz:
This would be amazing but I have a vague memory that for some types it was not known?

Or is that no longer true after I fixed GDAL...

What about Well-known geometry.jl ?

rafaqz:
Ah sorry I was confused with `trait`... would be nice to have that on the type

rafaqz (approved):

PR #186, opened 2025-01-20
Fix `UndefVarError` in `nfeature` and `getfeature`

simsurace:
`t` was not defined in these two functions, throwing an error e.g. on calling `GeoJSON.write` on a `Wrappers.FeatureCollection`.


simsurace:
I see. All the better!

PR #190, opened 2025-03-03
add 2d extent support

asinghvi17:


rafaqz:
☠️🦜⚓

rafaqz:
Needs to be in Makie

felixcremer:
Also this hard codes the names X and Y and would fail if your dimensions in the DimArray would be lat and lon and you extract an extent from that.

PR #194, opened 2025-05-15
Add a convert method for (module, trait, geom)

asinghvi17:
This allows the trait to be passed through completely from the time of discovery to the bottom level, which decreases allocations and dynamic dispatch in long call chains like GeometryOps' `apply`.

If this makes sense to add then I will add tests, docs, etc.

### DIFF
--- a/src/fallbacks.jl
+++ b/src/fallbacks.jl
@@ -148,7 +148,7 @@ calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filte
 # corresponding geometry type
 function convert(package::Module, geom)
     t = trait(geom)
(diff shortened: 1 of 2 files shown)


src/fallbacks.jl
```diff
@@ -148,6 +148,9 @@ calc_extent(t::AbstractFeatureCollectionTrait, fc) = reduce(Extents.union, filte
 # corresponding geometry type
 function convert(package::Module, geom)
     t = trait(geom)
+    convert(package, trait, geom)
```
asinghvi17, line 151:
```suggestion
    convert(package, t, geom)
```

rafaqz:
Makes sense to me

PR #195, opened 2025-05-20
Move MakieCore and GeometryBasics to an extension of GeoInterfaceMakie

rafaqz:
This PR moves MakieCore and GeometryBasics to an extension. Resolving https://github.com/JuliaGeo/Shapefile.jl/pull/139

Its a little more annoying than it could be because extending the macro in the extension seems to have world age issues - the extensions wont see the new methods added even if you can see them in the REPL before they load!

But, instead of putting the macro in an extension we can just keep it in the package and make a breaking change.

In the calling package we just add Makie or MakieCore manually:

```juila
GeoInterfaceMakie.@enable MakieCore Shapefile.AbstractShape
```

And it all seems to work fine, and MakieCore/GeometryBasics dependencies are now optional.

### DIFF
--- a/GeoInterfaceMakie/Project.toml
+++ b/GeoInterfaceMakie/Project.toml
@@ -23,6 +23,8 @@ julia = "1.6"
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 LibGEOS = "a90b1aa1-3769-5649-ba7e-abc5a9d163eb"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+MakieCore = "20f20a25-4f0e-4fdf-b5d1-57303727442b"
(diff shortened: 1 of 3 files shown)


evetion:
Can you explain why the two argument macro is now needed?

Note that in https://github.com/JuliaGeo/GeoInterface.jl/pull/143 we had an approach to rename and move the macros @enable_plots and @enable_makie to GeoInterface.jl stubs.

rafaqz:
We would need to move the Geometry basics GeoInterface extension into GeoInterface to avoid the dependency loop. But yes then your strategy can work.

But I think it will need the same double argument macro: the reason is it seems that we cant add methods to a macro in an extension that get used from another extension.

Adding a parameter for it let's the package just pass MacirCore it in so there is no dependencies needed for the macro part. Maybe you solved that another way but it seemed difficult to me.

I think it's just internal problem of how extending  macros works in extensions, I guess because they need to be loaded earlier than functions.

PR #197, opened 2025-07-01
support Makie v0.24

SimonDanisch:


asinghvi17:
Thanks!  If CI passes I think this is good to go

asinghvi17:
We should just finish up and merge that PR that removes GeoInterface from GeometryBasics and makes it an extension or something

PR #198, opened 2025-07-05
Move Makie and Plots recipes to extensions

rafaqz:
This PR removes the GeoInterfaceMakie and GeoInterfaceRecipes packages completely, and instead adds GeoInterfaceMakieExt and GeoInterfaceRecipesBaseExt extensions.

In this PR the macros to add recipes to geometries are:

```julia
GeoInterface.@enable_makie Makie MyGeometry
GeoInterface.@enable_plots RecipesBase MyGeometry
```

Theyre tested on the GeoInterface wrapper geometries, which means those will now just plot without loading an extra package.

The macro needs makie/plots in the name now as `@enable` could be anything outside of a plottting context. 

It also needs the package as the first argument to avoid world age problems, so packages can just apply the macro from GeoInterface in their extensions on Makie/RecipesBase. It doesn't work moving the inner macro code to the extension where Makie/RecipesBase is available. 

Closes #197 
Closes #195 
Closes #143

We should make a minor version bump after this, and all packages depending on GeoInterfaceMakie/GeoInterfaceRecipes will need to be switched to the macros defined here.

@SimonDanisch if you want to review, this should solve dependency problems.


evetion:
This is great! Thanks for making the effort to revamp plotting. I will go through it later today. Small question regarding the world age, would `invokelatest` work for the enable macro?

asinghvi17 (approved):
This looks great to me and should solve our problems.  Thanks!

asinghvi17:
GeometryBasics new version is now out.

felixcremer:
Should we move the tests of the enable macro which lived in the GeoInterfaceRecipes to the GeoInterface tests?

rafaqz:
Yeah, Makie broke MultiPoint

SimonDanisch:
What would be  a good test for multipoint support? 
Just 
```julia
scatter(MultiPoint(rand(Point2f, 10), rand(Point2f, 10))
lines(MultiPoint(rand(Point2f, 10), rand(Point2f, 10))
```
?
I guess lines should connect the first array and the second, but no connection between them?

rafaqz:
But doesn't LineString or Polygon do that already if you plot a vector of them?

asinghvi17:
That kind of splitting doesn't natively exist yet but we should implement that.  It will also resolve some long standing bugs with geomakie which is a plus

rafaqz:
Ok, I guess I assumed there was a general abstraction of that poly pattern for any vector of geometries. So yeah, that's my feature request!

Geospatial geometries are all expected to be able to represent the spatial component of some feature/table row, which has other associated values you often want to display visuall. One point, a multipoint collection, a polygon, a multipolygon, a linestring etc, are all similar kinds of thing. So from a GeoInterface perspective we should aim for them all to behave identically, besides the detail of the actual geometric shape they make.

rafaqz:
This is good to merge, please hit the button

PR #211, opened 2026-01-04
Implement GeoInterface on Extent

### LINKED ISSUE #178: Extents should have RectangleTrait

evetion:
Fixes #178. Only for the 2D case though, I don't think we can do anything for 3D+, unless we randomly pick the minimum of the added dimensions.


asinghvi17:
I think this probably needs a bit more consideration because right now, a lot of methods in the ecosystem, specifically GeometryOps and Rasters, assume that the trait of an extent is nothing and dispatch based on that.

So any change here would have to be accompanied by a corresponding change there. It's not technically a breaking change, but I would argue that it's breaking enough that a lot of packages will need to retro-cap their GeoInterface compat on the registry..

